PR #3, opened 2019-09-05
support python3.4 by importing contextlib2 if import of contextlib fails

Solvik:
I've experienced some issues on older distributions with python 3.4
contextlib is standard starting python 3.5, there's a project called [contextlib2]([https://contextlib2.readthedocs.io/en/stable/) that seems to be maintained 

This PR works fine with the tests I've done

### DIFF
--- a/jsonargparse.py
+++ b/jsonargparse.py
@@ -1,4 +1,3 @@
-
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


mauvilsa:
@Solvik python 3.4 already reached its end of life, so I would be skeptical on maintaining compatibility with it. But since it is such a small change all is good. Thanks for the contribution!

Solvik:
Thanks!

And yes, that's a shame to have to maintain compatibility on EOL distrib with EOL python version..
Also have to `pip3 install typing` because typing got standard in Python 3.5 :)

Solvik:
@mauvilsa pypi still mentions ` Requires: Python >=3.5 ` so pip3 cannot install the package. Should you change that?

mauvilsa:
Done

PR #16, opened 2020-11-04
Fix parse_args return type

carmocca:
To avoid `Expected type 'Dict[str, Any]', got 'Namespace' instead ` when `parse_as_dict=True` is used.

Some functions use type hints in the signature/return, others in the docstring, others in both. I changed only the docstring here because it is what it had.

Thanks!


mauvilsa:
Merged it. Though I was thinking of removing all types from docstrings, adding them as proper type hints and using sphinx-autodoc-typehints so that they all appear in the api reference.

PR #71, opened 2021-08-04
Fix Mapping typehint check and add more root types

### LINKED ISSUE #69: Shouldn't signatures typehints support collections?
I have a subclass that takes an arg as input of type `Union[Iterable[MyClass], Mapping[str, MyClass]]` (`Union`, `Iterable`, and `Mapping` coming from `typing`).

I see it fails because it checks with `is_class_typehint`, and none of `Iterable` nor `Mapping` are "classes":

https://github.com/omni-us/jsonargparse/blob/e7ab0f3c2e687563a5f5e8eca8087f61b39977dd/jsonargparse/signatures.py#L329-L335

But shouldn't it support collections when checking the typehints? They seem to actually be supported when instantiating or calling the action (using `adapt_typehints`).

bryant1410:
Fix #69

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ v3.18.0 (2021-??-??)
 
 Added
 ^^^^^
-- Support for parsing ``Mapping`` and ``MutableMapping` types.
+- Support for parsing ``Mapping`` and ``MutableMapping`` types.
 - Support for parsing ``frozenset``, ``MutableSequence`` and ``MutableSet`` types.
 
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -15,6 +15,7 @@ v3.18.0 (2021-??-??)
 Added
 ^^^^^
 - Support for parsing ``Mapping`` and ``MutableMapping` types.
```
mauvilsa, line 17:
```suggestion
- Support for parsing ``Mapping`` and ``MutableMapping`` types.
```

mauvilsa (changes requested):
Looks good! Only one quote to fix which I think was a mistake in the previous pull request.

mauvilsa (approved):

PR #72, opened 2021-08-04
Fix enum serialization

bryant1410:


bryant1410:
Not sure I understand the failing test. Seems to fail in `master` as well.

I see it's Python 3.5. Have you considered dropping support for it? [It reached end-of-life last year](https://www.python.org/downloads/release/python-350/).

mauvilsa:
> Not sure I understand the failing test. Seems to fail in `master` as well.
> 
> I see it's Python 3.5. Have you considered dropping support for it? [It reached end-of-life last year](https://www.python.org/downloads/release/python-350/).

It fails because docstring-parse now does not support python 3.5 and they didn't update the minimum version supported. I will not drop support for 3.5 just yet. I need to fix it so that docstring-parse is not installed on python 3.5.

mauvilsa:
In which cases the enum serialization failed? Might be good to add a unit test to prevent regressions like https://github.com/omni-us/jsonargparse/blob/5e561943dd4dfc524c406d2c5d607b44c41e7405/jsonargparse_tests/signatures_tests.py#L400

bryant1410:
I saw this expression with no assignment, I thought it didn't make sense. But now that I see it, is the expression to check that the name is correct, but `val` should be kept because it's serialization and already a string? In that case, maybe we should change this and add a comment? So it's clearer.

mauvilsa:
> I saw this expression with no assignment, I thought it didn't make sense. But now that I see it, is the expression to check that the name is correct, but `val` should be kept because it's serialization and already a string? In that case, maybe we should change this and add a comment? So it's clearer.

Sure, lets add a comment.

`val = typehint[val]` would convert it from a string into an enum value, no? I am wondering why this change did not make any unit test fail. Might mean that we are missing some test.

bryant1410:
> `val = typehint[val]` would convert it from a string into an enum value, no? I am wondering why this change did not make any unit test fail. Might mean that we are missing some test.

Yeah, I see the same now.

Would you mind changing/fixing the tests as you're more familiarized? Feel free to use this branch.

mauvilsa:
> Would you mind changing/fixing the tests as you're more familiarized? Feel free to use this branch.

I will take a note to look at it. Maybe you can use this pull request to add a comment.

bryant1410:
> I will take a note to look at it. Maybe you can use this pull request to add a comment.

What if I make the comment change after you fix the tests? Because this change as it is can help us fix it.

Thinking it again, instead of a comment I was thinking maybe we can put `assert val in typehint`, which is self-explanatory.

mauvilsa:
> What if I make the comment change after you fix the tests? Because this change as it is can help us fix it.

Thi possible missing test is not a priority for me. Not sure when I will look at it.

> Thinking it again, instead of a comment I was thinking maybe we can put `assert val in typehint`, which is self-explanatory.

I understand that asserts are disabled when enabling some optimizations. So an assert here would not be good. It should be an if and then a raise statement.

PR #75, opened 2021-08-06
Fix linking for deep targets

### LINKED ISSUE #73: Argument linking doesn't work for deep-level targets
Example file:

```python
import jsonargparse
from jsonargparse import ActionConfigFile


class D:
    pass


class A:
    def __init__(self, d: D) -> None:
        self.d = d


class BSuper:
    pass


class BSub(BSuper):
    def __init__(self, a: A) -> None:
        self.a = a


class C:
    def fn(self) -> D:
        return D()


if __name__ == '__main__':
    import a  # Otherwise the superclass namespace would be `__main__` instead of `a` and `issubclass` check would fail.
    # See https://stackoverflow.com/a/38477315/1165181
    parser.add_argument("--config", action=ActionConfigFile)
    parser.add_subclass_arguments(a.BSuper, nested_key="b", required=True)
    parser.add_class_arguments(C, nested_key="c")
    parser.link_arguments("c", "b.init_args.a.d", compute_fn=C.fn, apply_on="instantiate")

    args = parser.parse_args(["--config", "a.yml"])
```

`a.yml`:

```yml
b:
  class_path: a.BSub
  init_args:
    a:
      class_path: a.A
c:
```

bryant1410:
This fixes #73 by adding support for linked targets that have many levels.

Python file that tests this behavior:

```python
import jsonargparse
from jsonargparse import ActionConfigFile


class D:
    pass


class A:
    def __init__(self, d: D) -> None:
        self.d = d


class BSuper:
    pass


class BSub(BSuper):
    def __init__(self, a: A) -> None:
        self.a = a


class C:
    def fn(self) -> D:
        return a.D()  # If I don't put `a`, it's gonna be `__main__`.


if __name__ == '__main__':
    import a  # Otherwise the superclass namespace would be `__main__` instead of `a` and `issubclass` check would fail.
    # See https://stackoverflow.com/a/38477315/1165181
    parser.add_argument("--config", action=ActionConfigFile)
    parser.add_subclass_arguments(a.BSuper, nested_key="b", required=True)
    parser.add_class_arguments(C, nested_key="c")
    parser.link_arguments("c", "b.init_args.a.init_args.d", compute_fn=C.fn, apply_on="instantiate")

    args = parser.parse_args(["--config", "a.yml"])

    config = parser.instantiate_classes(args)

    assert isinstance(config["b"].a.d, a.D)
```

`a.yml`:

```yml
b:
  class_path: a.BSub
  init_args:
    a:
      class_path: a.A
c: {}
```

It's a different issue, but it feels like the parser created in `adapt_class_type` should get some things from the action's parser such as `logger`. For it, this method (and `adapt_typehints`) could be converted into `ActionTypeHint` class methods, get its parser and pass the property, along with potential others.

Finally, this PR also _tries_ to add support for linked targets that are inside dicts, **but I didn't test it**. Still, wanted to get feedback on it. Maybe this part can be for another PR. Also, feels like other paths apart from dicts could be supported, such as lists, as I only added code for it in one of the `if` conditions but not the rest.

### DIFF
--- a/jsonargparse/typehints.py
+++ b/jsonargparse/typehints.py
@@ -454,18 +454,21 @@ def adapt_class_type(val_class, init_args, serialize, instantiate_classes, sub_a
     parser = ArgumentParser(error_handler=None, parse_as_dict=True)
     parser.add_class_arguments(val_class, **sub_add_kwargs)
 
+    # No need to re-create the linked arg but just "inform" the corresponding parser actions that it exists upstream.
     for target in sub_add_kwargs.get('linked_targets', []):
-        for split in ['.init_args.', '.']:
-            if split in target:
-                prefix, suffix = target.split(split, maxsplit=1)
+        split_index = target.find(".")
+        if split_index != -1:
+            split = ".init_args." if target[split_index:].startswith(".init_args.") else "."
 
-                action = next(a for a in parser._actions if a.dest == prefix)
+            parent_key, key = target.split(split, maxsplit=1)
 
-                sub_add_kwargs = getattr(action, 'sub_add_kwargs')
-                sub_add_kwargs.setdefault('linked_targets', set())
-                sub_add_kwargs['linked_targets'].add(suffix)
+            action = next(a for a in parser._actions if a.dest == parent_key)
 
-                break
+            sub_add_kwargs = getattr(action, 'sub_add_kwargs')
+            sub_add_kwargs.setdefault('linked_targets', set())
+            sub_add_kwargs['linked_targets'].add(key)
+
+            break
 
     if instantiate_classes:
         init_args = parser.instantiate_subclasses(init_args)
--- a/jsonargparse/typehints.py
+++ b/jsonargparse/typehints.py
@@ -453,6 +453,17 @@ def adapt_class_type(val_class, init_args, serialize, instantiate_classes, sub_a
     from .core import ArgumentParser
     parser = ArgumentParser(error_handler=None, parse_as_dict=True)
     parser.add_class_arguments(val_class, **sub_add_kwargs)
+
+    for target in sub_add_kwargs.get('linked_targets', []):
+        if "." in target:
+            prefix, suffix = target.split(".", maxsplit=1)
+
+            action = next(a for a in parser._actions if a.dest == prefix)
+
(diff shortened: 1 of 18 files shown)


mauvilsa (changes requested):
Please add at least one unit test that is able to prevent regressions from this fix.

mauvilsa:
> It's a different issue, but it feels like the parser created in `adapt_class_type` should get some things from the action's parser such as `logger`. For it, this method (and `adapt_typehints`) could be converted into `ActionTypeHint` class methods, get its parser and pass the property, along with potential others.
>
> Finally, this PR also _tries_ to add support for linked targets that are inside dicts, **but I didn't test it**. Still, wanted to get feedback on it. Maybe this part can be for another PR. Also, feels like other paths apart from dicts could be supported, such as lists, as I only added code for it in one of the `if` conditions but not the rest.

Please do split. It seems you have several ideas and it would be easier if things are kept separate. This pull request focus it only on deep targets.

jsonargparse/typehints.py
```diff
@@ -410,7 +410,16 @@ def adapt_typehints(val, typehint, serialize=False, instantiate_classes=False, s
...
                 val = {cast(k): v for k, v in val.items()}
             for k, v in val.items():
-                val[k] = adapt_typehints(v, subtypehints[1], **adapt_kwargs)
+                if "linked_targets" in adapt_kwargs["sub_add_kwargs"]:
+                    kwargs = adapt_kwargs.copy()
+                    sub_add_kwargs = kwargs["sub_add_kwargs"]
+                    sub_add_kwargs["linked_targets"] = {t[len(k + "."):] for t in sub_add_kwargs["linked_targets"]
+                                                        if t.startswith(k + ".")}
+                    sub_add_kwargs["linked_targets"] = {t[len("init_args."):] if t.startswith("init_args.") else t
+                                                        for t in sub_add_kwargs["linked_targets"]}
```
mauvilsa, line 419: I see that now there is a unit test. However these lines are not covered by the test. Have a look at https://codecov.io/gh/omni-us/jsonargparse/src/32e125a857f7a2888ab1db063862d527e98ed7fc/jsonargparse/typehints.py#

mauvilsa (changes requested):

mauvilsa (approved):
Looks good!

PR #90, opened 2021-09-15
Set `SUPPRESS` as the default value for `add_subclass_arguments`

### LINKED ISSUE #83: Config check failed with `add_subclass_arguments(required=False)`
```python

class A:
    ...

class B:
    ...

parser = ArgumentParser()
parser.add_subclass_arguments((A, B), nested_key="foo", required=False)
args = parser.parse_args()
```
produces:
```
scratch_12.py: error: Configuration check failed :: Parser key "foo": Value {} does not validate against any of the types in typing.Union[__main__.A, __main__.B]:
  - Type <class '__main__.A'> expects an str or a Dict with a class_path entry but got "{}"
  - Type <class '__main__.B'> expects an str or a Dict with a class_path entry but got "{}"
```

My assumption was that by setting `required=False`, `--foo` does not need to be passed.
And `args` would be an empty dictionary.

carmocca:
Fixes #83

This `if` will now work:

https://github.com/omni-us/jsonargparse/blob/dde80fb00f48e3a922001542c47f16c986df7fd2/jsonargparse/core.py#L951

### DIFF
--- a/jsonargparse/signatures.py
+++ b/jsonargparse/signatures.py
@@ -2,7 +2,7 @@
 
 import inspect
 import re
-from argparse import Namespace
+from argparse import Namespace, SUPPRESS
 from functools import wraps
 from typing import Any, Callable, List, Optional, Set, Tuple, Type, Union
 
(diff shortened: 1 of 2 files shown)


mauvilsa (approved):
I don't remember why I set the default to an empty dict. SUPPRESS does make more sense.

Issue #117, opened 2022-01-21
[Feature Request] Add include other config file ability to --config parse param


shenmishajing:
## Motivation
see this [issue](https://github.com/PyTorchLightning/pytorch-lightning/issues/11221) from pytorch lightning.

## Feature
Several keywords we have to support: `__base__`, `__delete__`, `__import__`, `change_item`, `pre_item` and `post_item`.

<!DOCTYPE html><figure class="md-table-fig" cid="n39" mdtype="table" style="box-sizing: border-box; margin: 1.2em 0px; overflow-x: auto; max-width: calc(100% + 16px); padding: 0px; cursor: default; caret-color: rgb(184, 191, 198); color: rgb(184, 191, 198); font-family: &quot;Helvetica Neue&quot;, Helvetica, Arial, &quot;Segoe UI Emoji&quot;, sans-serif; font-size: 16px; font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; background-color: rgb(54, 59, 64); text-decoration: none;">

keyword | value | effect
-- | -- | --
`__base__` | `str` or `list[str]`,(each `str` should be a relative path from cur cofig file) | Merge every config one by one, current last.
`__delete__` | `True` or `str\|int` or `list[str\|int]`,`True` for delete all keys from other config, `str\|int` only delete the specific key (for dict) or index (for list) | Delete some part of config from other.
`__import__` | Any | Just delete this, for convenience of reference in yaml
`change_item` | `list[[index, item]]`,used only when merge list | Add ability of merg list, change the `list[index]` from other to `item`
`insert_item` | `list[[index, item, (extend)]]`,used only when merge list | Add ability of merg list, insert iterm to the `list` at `index`, extend=True if insert a list of items
`pre_item` | `Any`or `list[Any]`,used only when merge list | Add ability of merg list, add the value in the start of the list from other to item
`post_item` | `Any`or `list[Any]`,used only when merge list | Add ability of merg list, add the value in the end of the list from other to item

</figure>

## Pseudo implementation of merging (assuming the config is dict instead of namespace)

```python
def deep_update(source, override):
    """
    Update a nested dictionary or similar mapping.
    Modify ``source`` in place.
    """
[...]

Issue #122, opened 2022-02-21
couldn't open import "*.libsonnet": no match locally or in the Jsonnet library paths.


tshu-w:
## Issue

Due to the extensive functionality of jsonnet (as mentioned in https://github.com/omni-us/jsonargparse/issues/117#issuecomment-1044191711), I tried switching the default configuration to jsonnet format, but I ran into problems with import.

## Step to reproduce

Here is the minimal config to reproduce:
my directory structure:
```
.
├── conf
│   ├── name.libsonnet
│   └── test.jsonnet
└── test.py
```
`test.py`:
```python
from jsonargparse import CLI

def command(
    name: str = "Lucky",
    prize: int = 100
):
    """
    Args:
        name: Name of winner.
        prize: Amount won.
    """
    print(f'{name} won {prize}€!')

if __name__ == '__main__':
    CLI(parser_mode="jsonnet")
```
`conf/name.libsonnet`:
```
"Mike"
```
`conf/test.jsonnet`:
```
local name = import 'name.libsonnet';

{
  "name": name,
  "prize": 100,
}
```

If I run `python test.py --config conf/test.jsonnet`, I got the following error:

<details>
  <summary>Click to expand</summary>
    
  ```
  Traceback (most recent call last):
    File "/home1/wts/lightning-template/debug/test.py", line 15, in <module>
      CLI(parser_mode="jsonnet")
    File "/home1/wts/.local/share/conda/envs/template/lib/python3.9/site-packages/jsonargparse/cli.py", line 69, in CLI
      cfg = parser.parse_args(args)
    File "/home1/wts/.local/share/conda/envs/template/lib/python3.9/site-packages/jsonargparse/deprecated.py", line 112, in patched_parse
      cfg = parse_method(*args, _skip_check=_skip_check, **kwargs)
    File "/home1/wts/.local/share/conda/envs/template/lib/python3.9/site-packages/jsonargparse/core.py", line 345, in parse_args
      cfg, unk = self.parse_known_args(args=args)
    File "/home1/wts/.local/share/conda/envs/template/lib/python3.9/site-packages/jsonargparse/core.py", line 230, in parse_known_args
      namespace, args = self._parse_known_args(args, namespace)
    File "/home1/wts/.local/share/conda/envs/template/lib/python3.9/argparse.py", line 2062, in _parse_known_args
      start_index = consume_optional(start_index)
[...]

PR #124, opened 2022-02-22
Add include other config file ability in yaml parse mode

shenmishajing:
see #117

### DIFF
--- a/jsonargparse/core.py
+++ b/jsonargparse/core.py
@@ -13,6 +13,7 @@ from unittest.mock import patch
 
 from .formatters import DefaultHelpFormatter, empty_help, formatter_context, get_env_var
 from .jsonnet import ActionJsonnet
+from .yaml import parse_str
 from .jsonschema import ActionJsonSchema
 from .loaders_dumpers import check_valid_dump_format, dump_using_format, get_loader_exceptions, loaders, load_value, load_value_context
 from .namespace import is_meta_key, Namespace, split_key, split_key_leaf, strip_meta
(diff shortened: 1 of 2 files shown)


mauvilsa:
@shenmishajing thank you for creating a pull request. I have been following closely the discussions in https://github.com/omni-us/jsonargparse/issues/117 and I am aware of desires and needs from people. However, there are more aspects that I need to consider. For the moment this feature will not be added just yet, i.e. this pull request will not be merged any time soon.

For now I do want to comment one detail and hope that you modify the code accordingly. With few modifications this can be a custom loader, see https://jsonargparse.readthedocs.io/en/stable/#custom-loaders. There should be no modifications in file [core.py](https://github.com/omni-us/jsonargparse/pull/124/files#diff-60bc15f66c4fb3ac46be4ddce042192460826a98146bc5c17fb291b9b653f12d). The custom loader can then be registered as a new parser mode, e.g. `set_loader('yaml_custom', custom_yaml_load)` or it could replace the default yaml loader, i.e. `set_loader('yaml', custom_yaml_load)`.

jsonargparse/core.py
```diff
@@ -589,7 +590,10 @@ def _load_config_parser_mode(
             _jsonnet = import_jsonnet('_load_config_parser_mode')
             cfg_str = _jsonnet.evaluate_snippet(cfg_path, cfg_str, ext_vars=ext_vars, ext_codes=ext_codes)
         try:
-            cfg_dict = load_value(cfg_str)
+            if self.parser_mode == 'jsonnet':
+                cfg_dict = load_value(cfg_str)
+            else:
+                cfg_dict = parse_str(cfg_str, cfg_path = cfg_path)
```
tshu-w, line 596: I don't think introducing too much non-standard syntax by default is beneficial for long-term project maintenance. You can easily extend it with [custom loader](https://jsonargparse.readthedocs.io/en/stable/#custom-loaders).

Another thing to note is that since jsonargparse can automatically switch current working directory  to the config file path as discussed in https://github.com/omni-us/jsonargparse/issues/122, custom loader does not need to pass in the path to support relative path introdction.

shenmishajing:
Change this into a custom loader, every one can get the code from #117

PR #138, opened 2022-05-23
Support list append from command line and config file

### LINKED ISSUE #85: Support passing lists of class_path items via command line
When one wants to pass a list of objects to a parser, this works very well using configs:

```python
from typing import List, Optional

from jsonargparse import ActionConfigFile

from pytorch_lightning.utilities.cli import ArgumentParser


class A:
    def __init__(self, name: str):
        self.name = name

    def __repr__(self):
        return self.name

class B:
    def __init__(self, people: Optional[List[A]] = None):
        self.people = people

    def __repr__(self):
        return repr(list(self.people))


parser = ArgumentParser(parse_as_dict=True)
parser.add_argument("--config", action=ActionConfigFile)
parser.add_class_arguments(B, nested_key="foo")
args = parser.parse_args()
args = parser.instantiate_classes(args)

print(args)
```

```
foo:
  people:
    - class_path: __main__.A
      init_args:
        name: one
    - class_path: __main__.A
      init_args:
        name: two
    - class_path: __main__.A
      init_args:
        name: three
```
(ignore the silly names)

```python
python script.py --config=config.yaml
{'foo': [one, two, three]}
```

However, for users who do not want or cannot use configs, their only option is to write the entire config through command line:

```python
$ python script.py --foo.people='[{class_path: __main__.A, init_args: {name: one}}, {class_path: __main__.A, init_args: {name: two}}, {class_path: __main__.A, init_args: {name: three}}]'
{'foo': [one, two, three]}
```

This is very long to write and error prone.

## Pitch

Support the previous by doing:

```python
python script.py \
    --foo.people.class_path=__main__.A --foo.people.init_args.name=one \
    --foo.people.class_path=__main__.A --foo.people.init_args.name=two \
    --foo.people.class_path=__main__.A --foo.people.init_args.name=three
```

where the order is important, as the `init_args` apply to the previous item with the same nested key.

> Note: after #84 is implemented, this would become:
> ```python
> python script.py \
>    --foo.people=A --foo.people.name=one \
>    --foo.people=A --foo.people.name=two \
>    --foo.people=A --foo.people.name=three
> ```

## Additional context:

We are implementing this in https://github.com/PyTorchLightning/pytorch-lightning/pull/8815

mauvilsa:
Closes #85.

### DIFF
--- a/jsonargparse_tests/typehints_tests.py
+++ b/jsonargparse_tests/typehints_tests.py
@@ -8,6 +8,7 @@ import re
 import sys
 import unittest
 import uuid
+import warnings
 import yaml
 from calendar import Calendar
 from contextlib import redirect_stderr, redirect_stdout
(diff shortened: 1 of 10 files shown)


mauvilsa:
@carmocca can you please review. Mostly to look at the unit tests in case you observe any important use cases that I forgot to add.

jsonargparse_tests/typehints_tests.py
```diff
@@ -226,6 +227,44 @@ def test_complex_number(self):
...
+            self.assertIn('Replacing list value "[1, 2]" with "3"', str(w[0].message))
+
+
+    def test_list_append_config(self):
+        parser = ArgumentParser(error_handler=None)
+        parser.add_argument('--cfg', action=ActionConfigFile)
+        parser.add_argument('--val', type=List[int], default=[1, 2])
+        self.assertEqual([3, 4], parser.parse_args(['--cfg', 'val: [3, 4]']).val)
+        self.assertEqual([1, 2, 3], parser.parse_args(['--cfg', 'val+: 3']).val)
+        self.assertEqual([1, 2, 3, 4], parser.parse_args(['--cfg', 'val+: [3, 4]']).val)
```
carmocca, line 248: You could also add tests that interleave setting init_args as in the tests removed in https://github.com/PyTorchLightning/pytorch-lightning/pull/13129

carmocca (approved):
Love this!

PR #145, opened 2022-07-07
Relax constraint on env_prefix when reading from environment

mougams:
Reading parameters from environment is a great feature, but user
can sometimes want to not have any prefix when reading environement
variables.
This patch introduces another parameter named empty_env_prefix,
that if set to True, will initialize env_prefix to None, and hence unlock
what has been described above.

### DIFF
--- a/jsonargparse/actions.py
+++ b/jsonargparse/actions.py
@@ -817,7 +817,7 @@ def parent_parsers_context(key, parser):
 class _ActionSubCommands(_SubParsersAction):
     """Extension of argparse._SubParsersAction to modify subcommands functionality."""
 
-    _env_prefix: Optional[str] = None
+    _env_prefix: Union[bool, str] = True
 
 
     def add_parser(self, name, **kwargs):
--- a/jsonargparse/core.py
+++ b/jsonargparse/core.py
@@ -162,7 +162,7 @@ class ArgumentParser(ActionsContainer, argparse.ArgumentParser):
     def __init__(
         self,
         *args,
-        env_prefix: Optional[str] = None,
+        env_prefix: Union[bool, str] = True,
         error_handler: Optional[Callable[['ArgumentParser', str], None]] = usage_and_exit_error_handler,
         formatter_class: Type[DefaultHelpFormatter] = DefaultHelpFormatter,
         logger: Union[bool, str, dict, logging.Logger] = False,
@@ -1390,7 +1390,7 @@ class ArgumentParser(ActionsContainer, argparse.ArgumentParser):
 
 
     @property
-    def env_prefix(self) -> Optional[str]:
+    def env_prefix(self) -> Union[bool, str]:
         """The environment variables prefix property.
 
         :getter: Returns the current environment variables prefix.
@@ -1399,12 +1399,17 @@ class ArgumentParser(ActionsContainer, argparse.ArgumentParser):
         Raises:
             ValueError: If an invalid value is given.
         """
-        return self._env_prefix
+        return self._env_prefix  # type: ignore
 
 
     @env_prefix.setter
-    def env_prefix(self, env_prefix: Optional[str]):
-        if env_prefix is None:
+    def env_prefix(self, env_prefix: Union[bool, str]):
+        if env_prefix is False:
+            self._env_prefix = None
+        elif env_prefix in {True, None}:
+            if env_prefix is None:
+                from .deprecated import deprecation_warning, env_prefix_property_none_message
+                deprecation_warning(ArgumentParser, env_prefix_property_none_message)
             self._env_prefix = os.path.splitext(self.prog)[0]
         elif isinstance(env_prefix, str):
             self._env_prefix = env_prefix
--- a/jsonargparse/deprecated.py
+++ b/jsonargparse/deprecated.py
@@ -250,3 +250,8 @@ logger_property_none_message = """
     Setting the logger property to None was deprecated in v4.10.0 and will raise
     an exception in v5.0.0. Use False instead.
 """
+
+env_prefix_property_none_message = """
+    Setting the env_prefix property to None was deprecated in v4.11.0 and will raise
+    an exception in v5.0.0. Use True instead.
+"""
--- a/jsonargparse_tests/test_core.py
+++ b/jsonargparse_tests/test_core.py
@@ -166,6 +166,24 @@ class ParsersTests(TempDirTestCase):
             self.assertFalse(parser.default_env)
 
 
+    def test_env_prefix(self):
+        parser = ArgumentParser(env_prefix=True, default_env=True, error_handler=None)
+        parser.add_argument("--test_arg", type=str, required=True, help="Test argument")
+        with self.assertRaises(ParserError):
+            with unittest.mock.patch.dict(os.environ, {'TEST_ARG': 'one'}):
+                parser.parse_args([])
+        prefix = os.path.splitext(parser.prog)[0].upper()
+        with unittest.mock.patch.dict(os.environ, {f'{prefix}_TEST_ARG': 'one'}):
+            cfg = parser.parse_args([])
+            self.assertEqual('one', cfg.test_arg)
+
+        parser = ArgumentParser(env_prefix=False, default_env=True)
+        parser.add_argument("--test_arg", type=str, required=True, help="Test argument")
+        with unittest.mock.patch.dict(os.environ, {'TEST_ARG': 'one'}):
+            cfg = parser.parse_args([])
+            self.assertEqual('one', cfg.test_arg)
+
+
     def test_parse_string(self):
         parser = example_parser()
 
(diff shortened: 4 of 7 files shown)


mauvilsa (changes requested):
Also please add a `v4.11.0 (2022-07-??)` section in the changelog with two entries, one in `Added` describing the new feature and another like https://github.com/omni-us/jsonargparse/blob/593dd785c36017f0a04ba1fc1fd403062c140dc6/CHANGELOG.rst#L84-L86

jsonargparse/core.py
```diff
@@ -1404,7 +1407,9 @@ def env_prefix(self) -> Optional[str]:
 
     @env_prefix.setter
     def env_prefix(self, env_prefix: Optional[str]):
```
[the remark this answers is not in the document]
mougams: I've updated to Union[bool, str, None] as it could also be Optional.

jsonargparse/core.py
```diff
@@ -1404,7 +1407,9 @@ def env_prefix(self) -> Optional[str]:
...
-        if env_prefix is None:
+        if self.empty_env_prefix:
+            self._env_prefix = None
+        elif env_prefix is None:
```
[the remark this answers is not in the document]
mougams: Done

jsonargparse_tests/test_actions.py
```diff
@@ -53,6 +53,10 @@ def test_ActionYesNo_parse_env(self):
         parser.add_argument('--op', action=ActionYesNo, default=False)
         self.assertEqual(True, parser.parse_env({'APP_OP': 'true'}).op)
 
+        parser = ArgumentParser(default_env=True, env_prefix=False)
```
mauvilsa, line 56: Please make the test standalone and don't use an action. The test should be in `test_core.py` since parsing of environment variables has nothing to do with actions.

Also please add a test for the deprecation message in `test_deprecated.py`.

jsonargparse/deprecated.py
```diff
@@ -250,3 +250,8 @@ def set_url_support(enabled:bool):
     Setting the logger property to None was deprecated in v4.10.0 and will raise
     an exception in v5.0.0. Use False instead.
 """
+
+env_prefix_property_none_message = """
+    Setting the env_prefix property to None was deprecated in v4.11.0 and will raise
+    an exception in v5.0.0. Use False instead.
```
mauvilsa, line 256: Should be "Use True instead" since that is what will have the same behavior as before with None.

mauvilsa (changes requested):

jsonargparse_tests/test_actions.py
mougams: Thanks for the pointer - done

jsonargparse/deprecated.py
mougams: Done

jsonargparse/actions.py
```diff
@@ -817,7 +817,7 @@ def parent_parsers_context(key, parser):
 class _ActionSubCommands(_SubParsersAction):
     """Extension of argparse._SubParsersAction to modify subcommands functionality."""
 
-    _env_prefix: Optional[str] = None
+    _env_prefix: Union[bool, str] = None
```
mauvilsa, line 820: Default True.

jsonargparse_tests/test_core.py
```diff
@@ -167,6 +167,23 @@ def test_default_env(self):
             self.assertFalse(parser.default_env)
 
 
+    def test_env_prefix(self):
+        for env_prefix in [None, True]:
```
mauvilsa, line 171: Don't test for `None`. The test should be such that no change would be needed when in v5.0.0 giving `env_prefix=None` raises an error.

mauvilsa (changes requested):

jsonargparse_tests/test_deprecated.py
```diff
@@ -155,6 +155,14 @@ def test_logger_property_none(self):
             LoggerProperty(logger=None)
 
 
+    @unittest.mock.patch.dict(os.environ, {"JSONARGPARSE_DEPRECATION_WARNINGS": "all"})
+    def test_env_prefix_none(self):
+        with warnings.catch_warnings(record=True) as w:
+            parser = ArgumentParser(env_prefix=None)
+            self.assertEquals(1, len(w))
+            self.assertIn('env_prefix', str(w[0].message))
```
mauvilsa, line 163: With your test I noticed that the other deprecation warnings were not really working. I did some changes, so please rebase master and look at the changed tests. Please use `str(w[-1].message)`.

jsonargparse_tests/test_deprecated.py
```diff
@@ -155,6 +155,14 @@ def test_logger_property_none(self):
             LoggerProperty(logger=None)
 
 
+    @unittest.mock.patch.dict(os.environ, {"JSONARGPARSE_DEPRECATION_WARNINGS": "all"})
+    def test_env_prefix_none(self):
+        with warnings.catch_warnings(record=True) as w:
+            parser = ArgumentParser(env_prefix=None)
+            self.assertEquals(1, len(w))
```
mauvilsa, line 162: Don't test the number of warnings.

jsonargparse_tests/test_deprecated.py
```diff
@@ -155,6 +155,14 @@ def test_logger_property_none(self):
             LoggerProperty(logger=None)
 
 
+    @unittest.mock.patch.dict(os.environ, {"JSONARGPARSE_DEPRECATION_WARNINGS": "all"})
```
mauvilsa, line 158: Remove this. The only test in which `env_prefix=None` should be this one, so only here we would see this warning and there is no need to enable all warnings.

jsonargparse_tests/test_core.py
```diff
@@ -167,6 +167,23 @@ def test_default_env(self):
...
+            parser = ArgumentParser(env_prefix=env_prefix, default_env=True)
+            parser.add_argument("--test_arg", type=str, required=True, help="Test argument")
+            with self.assertRaises(SystemExit):
```
mauvilsa, line 174: Please make sure that parsing error messages are not seen in the test logs. Options are:
- Instantiate `ArgumentParser` with `error_handler=None` and assert `ParserError` instead of `SystemExit`
- In this line add one more context manager `, suppress_stderr():`

jsonargparse_tests/test_core.py
```diff
@@ -167,6 +167,23 @@ def test_default_env(self):
...
+                with unittest.mock.patch.dict(os.environ, {'TEST_ARG': 'one'}):
+                    parser.parse_args([])
+            prefix = os.path.splitext(parser.prog)[0].upper()
+            with unittest.mock.patch.dict(os.environ, {f'{prefix}_TEST_ARG': 'one'}):
+                parser.parse_args([])
```
mauvilsa, line 179: Test that the parsed value is `'one'`.

jsonargparse_tests/test_core.py
```diff
@@ -167,6 +167,23 @@ def test_default_env(self):
...
+
+        parser = ArgumentParser(env_prefix=False, default_env=True)
+        parser.add_argument("--test_arg", type=str, required=True, help="Test argument")
+        with unittest.mock.patch.dict(os.environ, {'TEST_ARG': 'one'}):
+            parser.parse_args([])
```
mauvilsa, line 184: Test that the parsed value is `'one'`.

mauvilsa (changes requested):

jsonargparse_tests/test_core.py
mougams: Done

jsonargparse_tests/test_core.py
mougams: Done

jsonargparse_tests/test_core.py
mougams: Got it, done

jsonargparse_tests/test_deprecated.py
mougams: I noticed that I needed this when running the tests in my local venv. Seems ok now

jsonargparse_tests/test_deprecated.py
mougams: Removed

jsonargparse_tests/test_deprecated.py
mougams: I was update to suggest in another PR the change on this, as I noticed that the logger property test was really testing anything. Thx for the refacto

jsonargparse_tests/test_core.py
mougams: Removed

jsonargparse/actions.py
mougams: Done

mougams:
All tests/checks pass. But I'm not sure to understand why Windows 2019 test fail with

 `ModuleNotFoundError: No module named 'jsonargparse'`

mauvilsa (approved):
Now all looks good. The failing tests are due to the changes I did and not related to this pull request.

PR #179, opened 2022-10-21
Fix IPython AttributeError in parse_known_args

Erotemic:
## What does this PR do?

This is a minor fix. When running `parser.parse_known_args` in IPython, the inspect logic raises an AttributeError. This obfuscates the expected NotImplemented error message. Here is a MWE:

Step 1: Start IPython:

Step 2: Paste in:

```python
    import jsonargparse
    parser.add_argument('--foo')
    parser.parse_known_args(args=[])
```

Before the patch this causes `AttributeError: 'NoneType' object has no attribute '__package__'`. 

After this patch it correctly raises `NotImplementedError: parse_known_args not implemented to dissuade its use, since typos in configs would go unnoticed.`.

Now, I have opinions about being not being allowed to use `parse_known_args`. I'm not afraid of the potential shoegun, and often I need the functionality, but I'll save that for a different thread, this change should be fairly uncontroversial.

## Before submitting

- [na] Did you update **the documentation**? (readme and public docstrings)
- [na] Did you verify that new and existing **tests pass locally**?
- [na] Did you make sure that all changes preserve **backward compatibility**?
- [na] Did you update **the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

### DIFF
--- a/jsonargparse/core.py
+++ b/jsonargparse/core.py
@@ -224,7 +224,8 @@ class ArgumentParser(ActionsContainer, ArgumentLinking, argparse.ArgumentParser)
 
     def parse_known_args(self, args=None, namespace=None):
         """Raises NotImplementedError to dissuade its use, since typos in configs would go unnoticed."""
-        caller = inspect.getmodule(inspect.stack()[1][0]).__package__
+        caller_mod = inspect.getmodule(inspect.stack()[1][0])
+        caller = None if caller_mod is None else caller_mod.__package__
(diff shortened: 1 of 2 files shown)


mauvilsa:
Thanks for the pull request! Could you please add a unit test for this? It would be simple, just mock `inspect.module` to return `None` and assert that `parse_known_args` raises `NotImplementedError`.

Erotemic:
done

PR #197, opened 2022-11-23
Fix parsing failures for registered types that can't be cast to boolean

### LINKED ISSUE #196: Support for conversion to `torch.Tensor`
## 🐛 Bug report

I am encountering a problem using the package with Lightning.

When trying to specify an `init_args` which expects a `torch.Tensor` (for example when specifying the weights of a loss function), an error occurs because torch module expects a tensor, not a list.

It should be possible to [register](https://jsonargparse.readthedocs.io/en/stable/index.html#registering-types) pytorch tensors using

```python
def serializer(x):
    return x.tolist()

def deserializer(x):
    return torch.tensor(x)

register_type(torch.Tensor, serializer, deserializer)
```

However, doing so results in an error:

```bash
main: error: Parser key "model": Problem with given class_path "MyModel":
  - Configuration check failed :: Parser key "my_model": Value "Namespace(class_path='MySubModel', init_args=Namespace(..., loss=Namespace(class_path='torch.nn.CrossEntropyLoss', init_args=Namespace(weight=tensor([  7.6200, 104.9100,   1.5600,  20.6500,  14.0500,  14.0600,   1.0000,  36.0700]), size_average=None, ignore_index=-100, reduce=None, reduction='mean', label_smoothing=0.0))))" does not validate against any of the types in typing.Optional[MySubModel]:
    - Boolean value of Tensor with more than one value is ambiguous
...
```

Since a list is specified in the yaml and the namespace shows a `tensor([...])`, it looks like the deserializer is working correctly, but something is going wrong.

### Expected behavior

The list should be converted to a tensor without error.

### Environment

- jsonargparse 4.15.2
- Python 3.10.6

mauvilsa:
## What does this PR do?

Fixes #196

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ Fixed
 - ``parse_string`` raises ``AttributeError`` when given a simple string.
 - Added missing ``return_parser`` deprecation warning when ``CLI`` has
   subcommands.
+- Parsing fails for registered types that can't be cast to boolean `#196
+  <https://github.com/omni-us/jsonargparse/issues/196>`__.
 
 Changed
 ^^^^^^^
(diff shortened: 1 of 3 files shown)


samvanstroud:
Confirming this fixes #196, thanks @mauvilsa

PR #239, opened 2023-02-09
CLI support for callable class instances

### LINKED ISSUE #238: CLI does not work with callable component?
## 🐛 Bug report

CLI currently breaks if I pass in a callable instance instead of a function. It would be nice not to break this abstraction.

### To reproduce
```python

from jsonargparse import CLI


@dataclass
class Foo:
    def __call__(self, x: int):
        return x


if __name__ == "__main__":
    f = Foo()
    CLI(f, as_positional=False)
```

```
❯ python test.py                                
Traceback (most recent call last):
  File "/home/venky/dev/instant-science/name_generator/test.py", line 14, in <module>
    CLI(f, as_positional=False)
  File "/home/venky/dev/instant-science/name_generator/.venv/lib/python3.9/site-packages/jsonargparse/cli.py", line 69, in CLI
    _add_component_to_parser(component, parser, as_positional, fail_untyped, config_help)
  File "/home/venky/dev/instant-science/name_generator/.venv/lib/python3.9/site-packages/jsonargparse/cli.py", line 116, in _add_component_to_parser
    added_args = parser.add_class_arguments(component, **kwargs)
  File "/home/venky/dev/instant-science/name_generator/.venv/lib/python3.9/site-packages/jsonargparse/signatures.py", line 70, in add_class_arguments
    raise ValueError(f'Expected "theclass" parameter to be a class type, got: {theclass}.')
ValueError: Expected "theclass" parameter to be a class type, got: Foo().
```

### Expected behavior

It should work similar to `def foo(x: int)`.

### Environment

- jsonargparse version (e.g., 4.8.0): 4.19
- Python version (e.g., 3.9): 3.9
- How jsonargparse was installed (e.g. `pip install jsonargparse[all]`):  pip install jsonargparse[all]
- OS (e.g., Linux): linux

mauvilsa:
## What does this PR do?

Fixes #238

## Before submitting

### DIFF
--- a/jsonargparse/cli.py
+++ b/jsonargparse/cli.py
@@ -52,7 +52,10 @@ def CLI(
         module = inspect.getmodule(caller).__name__  # type: ignore
         components = [
             v for v in caller.f_locals.values()
-            if (inspect.isfunction(v) or inspect.isclass(v)) and inspect.getmodule(v).__name__ == module  # type: ignore
+            if (
+                (inspect.isclass(v) or callable(v)) and
+                getattr(inspect.getmodule(v), '__name__', None) == module
+            )
         ]
         if len(components) == 0:
             raise ValueError('Either components argument must be given or there must be at least one '
@@ -60,6 +63,10 @@ def CLI(
 
     elif not isinstance(components, list):
         components = [components]
+
+    unexpected = [c for c in components if not (inspect.isclass(c) or callable(c))]
+    if unexpected:
+        raise ValueError(f'Unexpected components, not class or function: {unexpected}')
 
     parser = parser_class(default_meta=False, **kwargs)
     parser.add_argument('--config', action=ActionConfigFile, help=config_help)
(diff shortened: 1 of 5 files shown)


jsonargparse/cli.py
```diff
@@ -124,12 +126,16 @@ def _add_component_to_parser(component, parser, as_positional, fail_untyped, con
             if not added_subargs:
                 remove_actions(subparser, (ActionConfigFile, _ActionPrintConfig))
             subcommands.add_subcommand(key, subparser, help=get_help_str(getattr(component, key), parser.logger))
+    else:
+        added_args = parser.add_function_arguments(component, as_group=False, **kwargs)
```
indigoviolet, line 130: you could also use `callable()` to verify that this is function-like

indigoviolet (commented):
looks great, thank you for the quick fix!

jsonargparse/cli.py
mauvilsa: I added a check before so at this point it will never be anything except a callable. No point in checking it again. Still changed the code because the check was an assert which wouldn't be executed if the code is optimized.

PR #242, opened 2023-02-15
Fix `fail_untyped` for Dataclasses

### LINKED ISSUE #241: Update `fail_untyped` flag to work with dataclasses whose arguments are classese with untyped arguments.
## 🚀 Feature request

The `fail_untyped` works with classes and nested classes, but id does not work with dataclasses which have class arguments with untyped arguments in their `__init__`.

### Motivation

Dataclasses come handy when we build complex CLI frameworks as it organise the input fields without the need of typing `class_path` and `init_args`. The powerful feature here is to have objects as arguments that could be configured directly from the CLI. However, even though when these objects have arguments that are typed it works great, but the majority of open source doesn't, and thus limits use quite a bit.

Here is a snip it to reproduce and to make sense of the above, quite complex 😅, description:
```py3
# main.py
from jsonargparse import CLI

class Database:
    def __init__(self, version) -> None:  # <-- HERE version is untyped; for `version: str` works!
        self._version = version

@dataclass
class MySQLConfig:
    db: Database
    host: str = "localhost"
    port: int = 8080

def my_app(cfg: MySQLConfig) -> None:
    print(f"{cfg.host}, {cfg.port}, {cfg.db._version}")

if __name__ == "__main__":
    CLI(my_app, fail_untyped=False)
```

Command:
```sh
python3 main.py --cfg.host "localhost" --cfg.port 8080 --cfg.db Database --cfg.db.version "0.0.1"
````
mind that when changing `version` to `version: str`, it works as expected.

Thank you for you time and for maintaining this great library!

### PR DESCRIPTION
**Pull request by ioangatop**
## What does this PR do?

Closes #241

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,9 +25,6 @@ Fixed
 - ``add_subcommands`` fails when parser has required argument and default config
   available `#232 <https://github.com/omni-us/jsonargparse/issues/232>`__.
 
-- ``fail_untyped`` does not work with dataclasses
-  available `#241 <https://github.com/omni-us/jsonargparse/issues/241>`__.
-
 Changed
 ^^^^^^^
 - When parsing fails now ``argparse.ArgumentError`` is raised instead of
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,7 @@ v4.19.0 (2022-12-27)
 
 Added
 ^^^^^
+- ``add_dataclass_arguments`` now supports the ``fail_untyped`` parameter
 - ``CLI`` now supports the ``fail_untyped`` and ``parser_class`` parameters.
 - ``bytes`` and ``bytearray`` registered on first use and decodes from standard
   Base64.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,7 +26,7 @@ Fixed
   available `#232 <https://github.com/omni-us/jsonargparse/issues/232>`__.
 
 - ``fail_untyped`` does not work with dataclasses
-  available `#241 <https://github.com/omni-us/jsonargparse/issues/241`__.
+  available `#241 <https://github.com/omni-us/jsonargparse/issues/241>`__.
 
 Changed
 ^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,9 @@ Fixed
 - ``add_subcommands`` fails when parser has required argument and default config
   available `#232 <https://github.com/omni-us/jsonargparse/issues/232>`__.
 
+- ``fail_untyped`` does not work with dataclasses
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by mauvilsa — CHANGES_REQUESTED** [review]
@ioangatop thank you for the effort to contribute!

From what I see this pull request adds the `fail_untyped` parameter to `add_dataclass_arguments`. Indeed this is missing and it seems to resolve #241.

However, I see there is no unit test added. For every bug fix and new feature at least one unit test should be added, even if the coverage does not decrease with the code changes. Please convert the code snippet in the description #241 into a unit test.

**Comment by ioangatop** [comment]
> @ioangatop thank you for the effort to contribute!
> 
> From what I see this pull request adds the `fail_untyped` parameter to `add_dataclass_arguments`. Indeed this is missing and it seems to resolve #241.
> 
> However, I see there is no unit test added. For every bug fix and new feature at least one unit test should be added, even if the coverage does not decrease with the code changes. Please convert the code snippet in the description #241 into a unit test.

@mauvilsa added!

File: CHANGELOG.rst [hunk 1]
```diff
@@ -25,6 +25,9 @@ Fixed
 - ``add_subcommands`` fails when parser has required argument and default config
   available `#232 <https://github.com/omni-us/jsonargparse/issues/232>`__.
 
+- ``fail_untyped`` does not work with dataclasses
+  available `#241 <https://github.com/omni-us/jsonargparse/issues/241`__.
```
mauvilsa, line 29: There is a missing `>` symbol.

ioangatop: done!

CHANGELOG.rst
```diff
@@ -25,6 +25,9 @@ Fixed
 - ``add_subcommands`` fails when parser has required argument and default config
   available `#232 <https://github.com/omni-us/jsonargparse/issues/232>`__.
 
+- ``fail_untyped`` does not work with dataclasses
```
mauvilsa, line 28: Also, please move this to the end of the `Added` section without leaving any space in between. And change it to something like ` ``add_dataclass_arguments`` now supports the ``fail_untyped`` parameter `.

ioangatop: Right, done!

mauvilsa: What I meant was to remove it from Fixed since it is not a bug fix. It is a new feature. Remove this item and move the link to the Added section.

ioangatop: sure, fixed!

mauvilsa (approved):

ioangatop:
@mauvilsa thank you very much for the review and merge! 

Question: because this feature is kinda important for us, do you know when you will tag a new version and release it?

Thank you again

mauvilsa:
To release I need to run further tests with projects where we use jsonargparse. I am currently traveling and unable to do that. I will try to look into it early next week.

Issue #244, opened 2023-02-19
Customized Class Signatures


Erotemic:
## 🚀 Feature request

I would like a way for a class to have programatic mechanism such that a developer can tell jsonargparse what it's signature is and so they have control over how its arguments are parsed.

I want to describe something I'm currently doing to extend jsonargparse that demonstrates this idea, and brainstorm how something like this would best be intergrated into jsonargparse. And if it is the case that this feature isn't desired, then I hope to at least make the maintainers aware of what I'm doing so maybe they are less likely to break my monkey patches.

### Motivation

Currently jsonargparse does a fine job of adding class arguments, provided those arguments are introspectable at runtime. I really like how it can peek into the docstring and extract useful help message. However, there are a few things I would like more control of:

* a mechanism for specifying the parameter name exactly once, and couple that with the documentation. 
* aliases and shortform names
* customized argparse actions.

### Pitch

I maintain yet another argument parsing library called [scriptconfig](https://pypi.org/project/scriptconfig/). It works similarly to `attrs`, but it focuses on giving a flat key/value configuration so a function can be invoked via kwargs, sys.argv, or a config file in the same way. It is declarative. You define a class, and the class variables assign metadata to attributes, which can then be instantiated / validated / used to make an argparse object / etc... Because all of the information about a variable is stored as a class attribute, you can couple the help docs with the declaration of the variable. You can also give each variable an alias, which helps a lot when you want to change a variable name, but you don't want to break existing users.

```python
import scriptconfig as scfg

class MyClassConfig(scfg.DataConfig):
    key1 = scfg.Value(1, alias=['key_one'], help='description1')
    key2 = scfg.Value(None, help='description2')
    key3 = scfg.Value(False, isflag=True, help='description3')
```

I use this to also define init signatures for complex classes that might have a lot of different - and evolving - arguments. This is particularly useful with pytorch lightning modules and datamodules, where I tend to use the following pattern to ensure everything is specified in a config dictionary. 

```python
class MyClass:

    def __init__(self, **kwargs):
        self.config = MyClassConfig(**kwargs)
```

I've recently switched to LightningCLI, and to ensure that my modules can be used with jsonargparse I've done something like this: 

```python
class MyClass:
    __scriptconfig__ = MyClassConfig

    def __init__(self, **kwargs):
        self.config = MyClassConfig(**kwargs)
```

[...]

PR #255, opened 2023-03-18
Allow configuration aliases

Erotemic:
## What does this PR do?

This is a proof of concept that allows for aliases in config files. I've found that it works in my use case, but I don't know if it properly fits into jsonargparse the way it is currently written. I would like guidance on how this feature might be sustainably integrated.

The current change is this: If a the user defines a parse action that has multiple option strings, the normalized form of those names is now a valid config key that the user can specify in a yaml config and it maps to the dest name of the corresponding action.

Thus if the underlying ArgumentParser has an action that could be specified on the commandline as `--foo=1` or `--bar=1`, then it can now also accept a configuration dictionary specified as `foo: 1` or `bar: 1`. Here is an example:

```
    import jsonargparse
    parser.add_argument('--foo', '--bar')
    parsed = parser.parse_string('foo: 3')
    print(parsed)
    parsed = parser.parse_string('bar: 2')
    print(parsed)
    parsed = parser.parse_args(['--bar', '4'])
    print(parsed)
```

Currently prints:

```
Namespace(foo=3)
Namespace(foo=2, bar=2)
Namespace(foo='4')
```

In the second case I could probably fix it so it doesn't populate the alias name as well and always uses the primary key, but I'll leave that as TODO until I get feedback.

This PR has to do with #244, where I use scriptconfig to help define these aliases. Having access to aliases is important to me because it makes it much easier to transition / update variable names. I can give users a grace period where the old and new name are treated equally before writing specialized deprecation code, and finally removing the alias. Sometimes aliases are useful as permanent structures because it lets you define a CLI that works with multiple people's intuitions. Of course aliases do have drawbacks, but I hope maintainers see enough value in this feature to help me refine and merge this PR.

## Before submitting

- [ ] Did you verify that new and existing **tests pass locally**?

I have not done all before submitting tasks because I need feedback before I can prepare something that is merge-ready.


mauvilsa:
Sounds good to add support for aliases in config files. Makes it more consistent when an argument has several option strings. Thank you for working on this. I will add comments on the code for guidance.

jsonargparse_tests/test_alias.py
```diff
@@ -0,0 +1,10 @@
+def test_alias():
```
mauvilsa, line 1: Note that this repo uses the standard unittest. Simply adding a function that starts with `test_` will not run as part of the tests.

jsonargparse_tests/test_alias.py
mauvilsa, line 1: The current organization of unit test is somewhat bad. However, please follow the current structure which mean not randomly adding a new test file. This feature seems to be core, so it would go in `test_core.py`.

mauvilsa, line 1: This feature will need more than one unit test function. Things that probably should be tested are:
- Successful parsing of a value in an alias saved as the `dest` key and the alias not appearing in the namespace.
- Failure to parse alias because the value does not agree with the argument type.
- Alias in a nested namespace.
- Alias with name that includes dashes `-`.
- Maybe more?

mauvilsa (changes requested):
Additional to the other comments, note that these changes have made other unit tests fail.

jsonargparse_tests/test_alias.py
Erotemic: Agreed. Slowly working on this.

jsonargparse/actions.py
```diff
@@ -70,8 +70,11 @@ def _find_action_and_subcommand(
     if exclude is not None:
         actions = [a for a in actions if not isinstance(a, exclude)]
     fallback_action = None
+
+    ALLOW_ALIAS = True
+
     for action in actions:
-        if action.dest == dest:
+        if action.dest == dest or (ALLOW_ALIAS and dest in get_alias_dest(action)):
```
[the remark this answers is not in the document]
mauvilsa: Oh, right. Only the actions defined in jsonargparse inherit from this class. However, there are normal argparse actions and `_StoreAction` is one of them. Patching the argparse Action class to have this property would not be nice, but can be an option. Another option is to make it a normal attribute instead of a property and have a function (say `add_aliases(action)`) to add it. This function would be called for example before this line
https://github.com/omni-us/jsonargparse/blob/927ec027f41520cc768d501387ea678a096848c8/jsonargparse/core.py#L134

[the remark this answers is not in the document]
Erotemic: Patching of the `argparse.Action` class seems a bit heavy handed and prone to breakage depending on import and instance creation order. 

Patching an aliases attribute in on an instance level in the place you mentioned seems more reasonable, but I'm wondering if it makes the code harder to read. For example, a reader sees this magic alias property used, but doesn't see where it comes from. On the other hand, if there is a `get_aliases` function (which checks for and then sets the attribute so it isn't recomputed), then it's clear where the attribute is coming from. Either of the two options will work though, so let me know if you have a preference.

mauvilsa:
#956 added support for aliases, so closing this one.

PR #281, opened 2023-05-05
drop redundant mypy check from pre-commit

Borda:
## What does this PR do?

drop `mypy` from pre-commit to make it run faster locally while it is already as separate check in Github CI

## Before submitting


mauvilsa:
@Borda thank you so much for your contributions! I am not sure about this one. I like knowing if there are typing issues before pushing. Also mypy is only a small part of the pre-commit run time, since the unit tests and doc tests are also run. Why do you think removing mypy locally would be important?

Borda:
> I like knowing if there are typing issues before pushing. Also mypy is only a small part of the pre-commit run time, since the unit tests and doc tests are also run.

Personally, I would say that the recommit is very heavy and takes quite a long time to run.. also the `mypy` creates own cache, which would take some space in your local FS... :rabbit: 

But you are right. I would also like to see those checks before Puch, but not for each commit if I work locally and just wan to keep my work structured... :chipmunk: 

Anyway, this was just a minor suggestion, nothing that would significantly affect contributors :)

PR #304, opened 2023-06-14
stop subclass dive if you hit bad import

Borda:
## What does this PR do?

Recently I was facing losing potentially broken packages (see: https://github.com/Lightning-AI/lightning/pull/17829) and did not find other way around then this one, but happy to learn what I should do otherwise :chipmunk: 

## Before submitting

### DIFF
--- a/jsonargparse/typehints.py
+++ b/jsonargparse/typehints.py
@@ -920,7 +920,8 @@ def get_all_subclass_paths(cls: Type) -> List[str]:
     def add_subclasses(cl):
         try:
             class_path = get_import_path(cl)
-        except (ImportError, AttributeError):  # Attribute is added in case of dot notation imports
+        except (ImportError, AttributeError) as err:  # Attribute is added in case of dot notation imports
+            warning(f"Hit failing import with following trace: {err}")
             return
         if is_local(cl) or issubclass(cl, LazyInitBaseClass):
             return
--- a/jsonargparse/typehints.py
+++ b/jsonargparse/typehints.py
@@ -4,6 +4,7 @@ import inspect
 import os
 import re
 import sys
+import traceback
 from argparse import ArgumentError
 from collections import abc, defaultdict
 from contextlib import contextmanager, suppress
@@ -920,8 +921,8 @@ def get_all_subclass_paths(cls: Type) -> List[str]:
     def add_subclasses(cl):
         try:
             class_path = get_import_path(cl)
-        except (ImportError, AttributeError) as err:  # Attribute is added in case of dot notation imports
-            warning(f"Hit failing import with following trace: {err}")
+        except (ImportError, AttributeError):  # Attribute is added in case of dot notation imports
+            warning(f"Hit failing import with following trace: {traceback.format_exc()}")
             return
         if is_local(cl) or issubclass(cl, LazyInitBaseClass):
             return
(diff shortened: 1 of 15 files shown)


mauvilsa (changes requested):
To me here seems to be the correct place to fix it. It is actually a bug, because the `--help` of a CLI would crash.

Please add a unit test for get_all_subclass_paths mocking that an import fails. Also please add an entry to the changelog.

jsonargparse/typehints.py
```diff
@@ -918,7 +918,10 @@ def is_private(class_path):
         return "._" in class_path
 
     def add_subclasses(cl):
-        class_path = get_import_path(cl)
+        try:
+            class_path = get_import_path(cl)
+        except (ImportError, AttributeError):  # Attribute is added in case of dot notation imports
+            return
```
Borda, line 924: @mauvilsa shall we also print the error here? :thinking:

Borda:
> Please add a unit test for get_all_subclass_paths mocking that an import fails

not sure how this mock properly as this with mock it calls `__original__` yielding in mock not in the mocked ImportError

mauvilsa:
> > Please add a unit test for get_all_subclass_paths mocking that an import fails
> 
> not sure how this mock properly as this with mock it calls `__original__` yielding in mock not in the mocked ImportError

Could be something like:

```python
from unittest.mock import patch
from jsonargparse.util import get_import_path

class ImportClass:
    pass

def test_get_all_subclass_paths_import_error():

    def mocked_get_import_path(cls):
        if cls is ImportClass:
            raise ImportError
        return get_import_path(cls)

    with patch("jsonargparse.typehints.get_import_path", mocked_get_import_path):
        subclass_paths = get_all_subclass_paths(ImportClass)
    assert subclass_paths == []
```

jsonargparse_tests/test_typehints.py
```diff
@@ -18,19 +18,22 @@
...
 )
+from unittest import mock
 
 import pytest
 import yaml
 
 from jsonargparse import ActionConfigFile, ArgumentError, Namespace, lazy_instance
-from jsonargparse.typehints import ActionTypeHint, Literal, is_optional
+from jsonargparse.typehints import ActionTypeHint, Literal, is_optional, resolve_class_path_by_name, \
+    get_all_subclass_paths
```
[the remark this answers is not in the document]
Borda:
```suggestion
from jsonargparse.typehints import ActionTypeHint, Literal, is_optional, \
    get_all_subclass_paths
```

jsonargparse/typehints.py
```diff
@@ -918,7 +918,11 @@ def is_private(class_path):
         return "._" in class_path
 
     def add_subclasses(cl):
-        class_path = get_import_path(cl)
+        try:
+            class_path = get_import_path(cl)
+        except (ImportError, AttributeError) as err:  # Attribute is added in case of dot notation imports
+            warning(f"Hit failing import with following trace: {err}")
```
mauvilsa, line 924:
```suggestion
            warning(f"Hit failing import with following error: {err}")
```

`{err}` is not a trace. It is only the message of the exception. The warning would show the line number here, but not a complete trace.

Borda: so use `traceback` or another tool?
btw, I tried just printing the err and it was quite fine...

mauvilsa: I don't think there is much benefit of printing the traceback here. Just the error message is good enough.

mauvilsa (approved):

PR #341, opened 2023-08-01
update _typehints error types for "Union"

HsunGong:
## What does this PR do?

During debugging, I found that this package will raise annoy error messages as below figure (1).
image

After applying my pull request, we distinguish the different error types as figure (2).
image

## Before submitting

- [*] Did you read the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)?
- [*] Did you verify that new and existing **tests pass locally**?
- [*] Did you make sure that all changes preserve **backward compatibility**?

### DIFF
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -556,7 +556,9 @@ class UnexceptedClassError(ValueError):
     pass
 
 
-def raise_instantiation_error(message: str, val: Any = inspect._empty, exception: Optional[Exception] = None) -> NoReturn:
+def raise_instantiation_error(
+    message: str, val: Any = inspect._empty, exception: Optional[Exception] = None
+) -> NoReturn:
     if val is not inspect._empty:
         message += f". Got value: {val}"
     raise UnexceptedClassError(message) from exception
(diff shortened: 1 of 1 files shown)


mauvilsa (changes requested):
@HsunGong thank you for your interest to contribute!

I guess you have seen that a bunch of unit test fail. Your changes make valid parsing use cases work incorrectly. Interestingly you checked the box saying that the tests ran successfully for you, but this is not possible.

Note that parsing errors are currently printed this way by design. If you want to propose a change, you should understand why they are like this. So let me explain. When there is a `Union`, any of the subtypes can be valid. An actual parse error should only happen if **all** the subtypes in the `Union` fail. And the errors for all subtypes must be shown, otherwise the user will not know what is the problem. Take for example a type like `Union[BaseClassA, BaseClassB, int]`. Just because a parsed value is invalid for `BaseClassA` it does not mean the parsing should fail, because it could be valid for `BaseClassB`.

In the description above you say that the errors are "annoying", but with those screenshots it is not clear why you think this. Please clearly explain what is the problem with the current errors. Then explain your proposed change, motivating why this is an improvement to the current one.

mauvilsa:
@HsunGong please respond to my comment.

In the screenshot where it says "Not a valid subclass of **None** Logger" is not useful and would be good to improve. Though would be nice to have a code snippet to reproduce it.

mauvilsa:
Closing this since there has been no response.

PR #383, opened 2023-09-15
fix `parse_args` for `Callable` with multiple input arguments

### LINKED ISSUE #372: class instantiation with multiple parameters
Hello, I have a question on how I can instantiate a class that requires multiple parameters that determined during run time only?
I followed the example on [Optimizer](https://jsonargparse.readthedocs.io/en/v4.24.0/#callable-type) but can't figure out how I can expand it to multiple parameters:

```python
# Example from the website
class Optimizer:
    def __init__(self, params: Iterable):
        self.params = params


class SGD(Optimizer):
    def __init__(self, params: Iterable, lr: float):
        super().__init__(params)
        self.lr = lr


if __name__ == '__main__':
    value = {
        "class_path": "SGD",
        "init_args": {
            "lr": 0.01,
        },
    }

    parser = ArgumentParser()
    parser.add_argument("--optimizer", type=Callable[[Iterable], Optimizer])
    cfg = parser.parse_args(["--optimizer", str(value)])
    init = parser.instantiate_classes(cfg)
    optimizer = init.optimizer([1, 2, 3])
```

Let's say I want to add one more parameter to `SGD(params, weight, lr)`. How will the flow change?

```python
class SGD(Optimizer):
    def __init__(self, params: Iterable, weight: float, lr: float):
...
parser.add_argument("--optimizer", type=Callable[[Iterable, float], Optimizer])
```
gives the following error:
```
usage: example.py [-h] [--optimizer OPTIMIZER]
example.py: error: Parser key "optimizer":
  Expected a dot import path string: SGD
```
Setting `class_path` to `__main__.SGD` produces another error:
```
usage: example.py [-h] [--optimizer OPTIMIZER]
example.py: error: Parser key "optimizer":
  Type typing.Callable[[typing.Iterable, float], __main__.Optimizer] expects a function or a callable class: Expected '__main__.SGD' to be a class that instantiates into callable or a subclass of False.. Got value: Namespace(class_path='__main__.SGD', init_args=Namespace(lr=0.01))

```

hadipash:
## What does this PR do?

Fixes #372.

## Before submitting

### DIFF
--- a/DOCUMENTATION.rst
+++ b/DOCUMENTATION.rst
@@ -941,10 +941,9 @@ example the classes:
 
 
     class SGD(Optimizer):
-        def __init__(self, params: Iterable, lr: float, momentum: float = 0.):
+        def __init__(self, params: Iterable, lr: float):
             super().__init__(params)
             self.lr = lr
-            self.momentum = momentum
 
 .. testcode:: callable
     :hide:
@@ -958,20 +957,26 @@ A possible parser and callable behavior would be:
     >>> value = {
     ...     "class_path": "SGD",
     ...     "init_args": {
-    ...         "momentum": 0.9,
+    ...         "lr": 0.01,
     ...     },
     ... }
 
-    >>> parser.add_argument("--optimizer", type=Callable[[Iterable, float], Optimizer])  # doctest: +IGNORE_RESULT
+    >>> parser.add_argument("--optimizer", type=Callable[[Iterable], Optimizer])  # doctest: +IGNORE_RESULT
     >>> cfg = parser.parse_args(["--optimizer", str(value)])
     >>> cfg.optimizer
-    Namespace(class_path='__main__.SGD', init_args=Namespace(momentum=0.9))
+    Namespace(class_path='__main__.SGD', init_args=Namespace(lr=0.01))
     >>> init = parser.instantiate_classes(cfg)
-    >>> optimizer = init.optimizer([1, 2, 3], 0.01)
+    >>> optimizer = init.optimizer([1, 2, 3])
     >>> isinstance(optimizer, SGD)
     True
-    >>> optimizer.params, optimizer.lr, optimizer.momentum
-    ([1, 2, 3], 0.01, 0.9)
+    >>> optimizer.params, optimizer.lr
+    ([1, 2, 3], 0.01)
+
+Multiple dependency injection is also supported and can be specified the same
+way with ``Callable`` type hinting. For example, for two ``Iterable``
+dependencies, you can use the following syntax:
+``Callable[[Iterable, Iterable], Type]``. Please be aware that the injected
+dependencies passed as positional arguments (that is, passed as ``*args``).
 
 .. note::
 
@@ -998,11 +1003,11 @@ Then a parser and behavior could be:
     >>> parser.add_class_arguments(Model, 'model')
     >>> cfg = parser.get_defaults()
     >>> cfg.model.optimizer
-    Namespace(class_path='__main__.SGD', init_args=Namespace(lr=0.05, momentum=0.0))
+    Namespace(class_path='__main__.SGD', init_args=Namespace(lr=0.05))
     >>> init = parser.instantiate_classes(cfg)
     >>> optimizer = init.model.optimizer([1, 2, 3])
-    >>> optimizer.params, optimizer.lr, optimizer.momentum
-    ([1, 2, 3], 0.05, 0.0)
+    >>> optimizer.params, optimizer.lr
+    ([1, 2, 3], 0.05)
 
 See :ref:`ast-resolver` for limitations of lambda defaults.
 
--- a/DOCUMENTATION.rst
+++ b/DOCUMENTATION.rst
@@ -941,9 +941,10 @@ example the classes:
 
 
     class SGD(Optimizer):
-        def __init__(self, params: Iterable, lr: float):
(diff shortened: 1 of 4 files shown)


hadipash:
@mauvilsa Indeed the problem was in this line:
https://github.com/omni-us/jsonargparse/blob/826d7863ab34654af0d73d8f42fb5768ce64f5f3/jsonargparse/_typehints.py#L891
I also wrote a test and expanded docs. Let me know if need to correct anything.

DOCUMENTATION.rst
```diff
@@ -957,20 +958,20 @@ A possible parser and callable behavior would be:
     >>> value = {
     ...     "class_path": "SGD",
     ...     "init_args": {
-    ...         "lr": 0.01,
+    ...         "momentum": 0.9,
     ...     },
     ... }
 
-    >>> parser.add_argument("--optimizer", type=Callable[[Iterable], Optimizer])  # doctest: +IGNORE_RESULT
+    >>> parser.add_argument("--optimizer", type=Callable[[Iterable, float], Optimizer])  # doctest: +IGNORE_RESULT
```
mauvilsa, line 965: The changes in this example have made it very misleading. The concept being introduced here is parameters that are only available after the class has been instantiated. For the case of optimizer, this is only `params`. The type here must be `Callable[[Iterable], Optimizer])`, otherwise it makes what is being explained more confusing. `SGD` can have both `lr` and `momentum`, but both of these should be given in the `init_args`.

Certainly there could be classes that have more than one parameters available after instantiation, which is what this pull request is fixing. If you want to explain that this is possible, do so after the `SGD` example. And the example shouldn't be with `Optimizer` since that is a known pytorch example that has only one such parameter.

DOCUMENTATION.rst
```diff
@@ -957,20 +958,20 @@ A possible parser and callable behavior would be:
     >>> value = {
     ...     "class_path": "SGD",
     ...     "init_args": {
-    ...         "lr": 0.01,
+    ...         "momentum": 0.9,
```
mauvilsa, line 961: `lr` should be here.

DOCUMENTATION.rst
```diff
@@ -957,20 +958,20 @@ A possible parser and callable behavior would be:
...
     ...     },
     ... }
 
-    >>> parser.add_argument("--optimizer", type=Callable[[Iterable], Optimizer])  # doctest: +IGNORE_RESULT
+    >>> parser.add_argument("--optimizer", type=Callable[[Iterable, float], Optimizer])  # doctest: +IGNORE_RESULT
     >>> cfg = parser.parse_args(["--optimizer", str(value)])
     >>> cfg.optimizer
-    Namespace(class_path='__main__.SGD', init_args=Namespace(lr=0.01))
+    Namespace(class_path='__main__.SGD', init_args=Namespace(momentum=0.9))
```
mauvilsa, line 968: `lr` should be here.

DOCUMENTATION.rst
```diff
@@ -957,20 +958,20 @@ A possible parser and callable behavior would be:
...
     >>> init = parser.instantiate_classes(cfg)
-    >>> optimizer = init.optimizer([1, 2, 3])
+    >>> optimizer = init.optimizer([1, 2, 3], 0.01)
```
mauvilsa, line 970: This line should not change.

jsonargparse_tests/test_typehints.py
```diff
@@ -693,6 +696,44 @@ def test_callable_args_return_type_class(parser, subtests):
             assert f"{__name__}.{name}" in help_str
 
 
+def test_callable_multiple_args_return_type_class(parser, subtests):
+    parser.add_argument("--optimizer", type=Callable[[List[float], float], Optimizer], default=SGD)
```
mauvilsa, line 700: This looks good! Since it is just for testing, then there is no real risk that people use this as reference to learn the concept.

mauvilsa (changes requested):
Thank you very much for your contribution! It is looking good. The only issue I see is the docs.

DOCUMENTATION.rst
mauvilsa: Could also be just a small paragraph explaining that it is possible and only give as example a type hint.

hadipash: Good point. Fixed docs by adding a small paragraph for clarification.

DOCUMENTATION.rst
```diff
@@ -972,6 +972,12 @@ A possible parser and callable behavior would be:
     >>> optimizer.params, optimizer.lr
     ([1, 2, 3], 0.01)
 
+Multiple dependency injection is also supported and can be specified the same
+way with ``Callable`` type hinting. For example, for two ``Iterable``
+dependencies, you can use the following syntax:
+``Callable[[Iterable, Iterable], Type]``. Please be aware that the injected
+dependencies passed as positional arguments (that is, passed as ``*args``).
```
mauvilsa, line 979:
```suggestion
Multiple parameters available after injection is also supported and can be specified the same
way with a ``Callable`` type hint. For example, for two ``Iterable``
parameters, you can use the following syntax:
``Callable[[Iterable, Iterable], Type]``. Please be aware that the parameters are passed as positional arguments, this means that the injected function would be called like ``function(value1, value2)``.
```

This is not multiple dependency injection. The entire callable is the dependency. Multiple here are parameters only available after injection.

hadipash: Fixed

mauvilsa (approved):

PR #474, opened 2024-03-17
Feature path std io

mquillot:
## What does this PR do?

Add support for "-" as value for Path class initialization so that user
  can ask to use standard input/output instead of file.

## Before submitting

- [ X ] Did you read the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)?
- [ X ] Did you update **the documentation**? (readme and public docstrings)
- [ X ] Did you write **unit tests** such that there is 100% coverage on related code? (required for bug fixes and new features)
- [ X ] Did you verify that new and existing **tests pass locally**?
- [ X ] Did you make sure that all changes preserve **backward compatibility**?
- [ X ] Did you update **the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,13 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
+v4.28.0 (2024-03-??)
+--------------------
+
+Added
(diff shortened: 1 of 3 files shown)


mquillot:
Ok, I think we must add cases in the tests.
For know, when we use mode "r" or "w" it will not work I think. I should add these tests also.

PR #480, opened 2024-03-28
Fixes for non-stdlib dataclass-like types

nkrishnaswami:
## What does this PR do?
This PR includes a few small, related fixes for non-stdlib (possibly annotated) dataclasses and 3.12-style type aliases:
* Exclude non-init fields from `attrs` class instantiation.
* Exclude non-init fields from Pydantic 2 dataclass instantiation.
* Fix initialization for Pydantic 2 dataclasses with default values/factories.
* Look at underlying type for `Annotated` types where needed.
* Look at target for new `TypeAliasType` where needed. (This and the prior are commonly used with FastAPI)
* Tests showing the above were broken and are now fixed.

## Before submitting

- [X] Did you read the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)?
- [X] Did you update **the documentation**? (readme and public docstrings)
- [X] Did you write **unit tests** such that there is 100% coverage on related code? (required for bug fixes and new features)
- [X] Did you verify that new and existing **tests pass locally**?
- [X] Did you make sure that all changes preserve **backward compatibility**?
- [X] Did you update **the CHANGELOG**? (not for typos, docs, test updates, or minor internal changes/refactors)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,10 +14,20 @@ paths are considered internals and can change in minor and patch releases.
 v4.28.1 (2024-0?-??)
 --------------------
 
+Added
+^^^^^
+- `TypeAliasType` support added (`#480
+  <https://github.com/omni-us/jsonargparse/issues/480>`__).
+
 Fixed
 ^^^^^
--  Failure to process ``Annotated`` dataclass members, and inclusion of
-   non-init fields in `attrs` and Pydantic dataclass-like instantiation.
+- Attrs and Pydantic 2 dataclasses with non-init fields fail to instantiate
+  (`#480 <https://github.com/omni-us/jsonargparse/issues/480>`__).
+- Default values/factories for Pydantic 2 dataclasses with `Field` initializers
+  are not right (`#480 <https://github.com/omni-us/jsonargparse/issues/480>`__).
+- `Annotated` fields in dataclass-likes (eg FastAPI types) resolve incorrectly
+  (`#480 <https://github.com/omni-us/jsonargparse/issues/480>`__).
+
 
 v4.28.0 (2024-03-??)
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,14 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
+v4.28.1 (2024-0?-??)
+--------------------
+
+Fixed
+^^^^^
+-  Failure to process ``Annotated`` dataclass members, and inclusion of
+   non-init fields in `attrs` and Pydantic dataclass-like instantiation.
+
 v4.28.0 (2024-03-??)
 --------------------
 
@@ -18,14 +26,6 @@ Added
 ^^^^^
 - Support for "-" as value for Path class initialization so that user
   can ask to use standard input/output instead of file.
-
-v4.27.8 (2024-0?-??)
---------------------
-
-Fixed
-^^^^^
--  Failure to process ``Annotated`` dataclass members, and inclusion of
-   non-init fields in `attrs` and Pydantic dataclass-like instantiation.
 
 v4.27.7 (2024-03-21)
 --------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,14 @@ The semantic versioning only considers the public API as described in
 paths are considered internals and can change in minor and patch releases.
 
 
+v4.27.8 (2024-0?-??)
+--------------------
+
+Fixed
+^^^^^
+-  Failure to process ``Annotated`` dataclass members, and inclusion of
+   non-init fields in `attrs` and Pydantic dataclass-like instantiation.
+
 v4.27.7 (2024-03-21)
 --------------------
 
--- a/jsonargparse/_parameter_resolvers.py
+++ b/jsonargparse/_parameter_resolvers.py
@@ -949,11 +949,11 @@ def get_parameters_from_pydantic_or_attrs(
     if pydantic_support:
         pydantic_model = is_pydantic_model(function_or_class)
         if pydantic_model == 1:
-            fields_iterator = function_or_class.__fields__.items()  # type: ignore[union-attr]
+            fields_iterator = function_or_class.__fields__.items()
             get_field_data = get_field_data_pydantic1_model
             is_init_field = is_init_field_default
         elif pydantic_model > 1:
-            fields_iterator = function_or_class.model_fields.items()  # type: ignore[union-attr]
+            fields_iterator = function_or_class.model_fields.items()
             get_field_data = get_field_data_pydantic2_model
             is_init_field = is_init_field_default
         elif dataclasses.is_dataclass(function_or_class) and hasattr(function_or_class, "__pydantic_fields__"):
--- a/jsonargparse/_parameter_resolvers.py
+++ b/jsonargparse/_parameter_resolvers.py
@@ -923,10 +923,6 @@ def get_field_data_attrs(field, name, doc_params):
     )
 
 
-def is_init_field_pydantic_model(field) -> bool:
-    return True
-
-
 def is_init_field_pydantic2_dataclass(field) -> bool:
     from pydantic.fields import FieldInfo
 
@@ -956,11 +952,11 @@ def get_parameters_from_pydantic_or_attrs(
         if pydantic_model == 1:
             fields_iterator = function_or_class.__fields__.items()
             get_field_data = get_field_data_pydantic1_model
-            is_init_field = is_init_field_pydantic_model
+            is_init_field = lambda _: True
         elif pydantic_model > 1:
             fields_iterator = function_or_class.model_fields.items()
             get_field_data = get_field_data_pydantic2_model
-            is_init_field = is_init_field_pydantic_model
+            is_init_field = lambda _: True
         elif dataclasses.is_dataclass(function_or_class) and hasattr(function_or_class, "__pydantic_fields__"):
             fields_iterator = dataclasses.fields(function_or_class)
             fields_iterator = {v.name: v for v in fields_iterator}.items()
--- a/jsonargparse/_parameter_resolvers.py
+++ b/jsonargparse/_parameter_resolvers.py
@@ -926,7 +926,7 @@ def get_field_data_attrs(field, name, doc_params):
 def is_init_field_pydantic2_dataclass(field) -> Optional[bool]:
     from pydantic.dataclasses import FieldInfo
 
-    if isinstance(field.default, FieldInfo) and hasattr(field.default, 'init'):
+    if isinstance(field.default, FieldInfo) and hasattr(field.default, "init"):
         return field.default.init is not False
     return field.init is not False
 
--- a/jsonargparse/_parameter_resolvers.py
+++ b/jsonargparse/_parameter_resolvers.py
@@ -869,7 +869,7 @@ def get_field_data_pydantic1_model(field, name, doc_params):
 
 
 def get_field_data_pydantic2_dataclass(field, name, doc_params):
-    from pydantic.dataclasses import FieldInfo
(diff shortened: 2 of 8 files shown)


jsonargparse/_parameter_resolvers.py
```diff
@@ -907,34 +944,40 @@ def get_parameters_from_pydantic_or_attrs(
...
             get_field_data = get_field_data_attrs
+            should_init_field = should_init_field_default
 
     if not fields_iterator or not get_field_data:
         return None
 
     params = []
     doc_params = parse_docs(function_or_class, None, logger)
     for name, field in fields_iterator:
+        if should_init_field(field) is False:
```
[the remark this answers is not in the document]
nkrishnaswami: Second-guessing; that means the predicate's name is wrong.  Fixing it.

nkrishnaswami:
Hmm, seeing errors on pydantic < 2.6; how far back do you need to support?

mauvilsa (changes requested):
@nkrishnaswami thank you for contributing! I haven't had time to look at this in detail, so I will comment more later. For the moment, I see that there is a test failing in `test-py310-pydantic-v1`. The changes introduced in pydantic v2 are big, which means that many people will still be using v1 for a long time. Please make sure that the test passes because there is no plan yet to drop support for pydantic v1.

nkrishnaswami:
> For the moment, I see that there is a test failing in `test-py310-pydantic-v1`. ... Please make sure that the test passes because there is no plan yet to drop support for pydantic v1.
@mauvilsa Gotcha. I was having trouble running the tests in my virtual env, but I found the commands to use in the circleci config; I'll get this fixed shortly!

mauvilsa:
> @mauvilsa Gotcha. I was having trouble running the tests in my virtual env, but I found the commands to use in the circleci config; I'll get this fixed shortly!

You don't need to look at the circleci config to figure out how to run tests. In the [CONTRIBUTING.rst guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst#pre-commit) it is recommended that you setup pre-commit. And the pydantic v1 tests are run as part of the pre-push hooks. Also there it is explained how to manually run them.

nkrishnaswami:
@mauvilsa got the tests and CI passing!  Thanks!

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -468,9 +474,111 @@ def test_add_class_final(parser):
...
+if annotated and pydantic_support > 1:
+    import pydantic
+
+    @pydantic.dataclasses.dataclass(frozen=True)
+    class InnerDataClass:
+        """DataClass description
+
+        Args:
+            a1: a1 help
+        """
```
mauvilsa, line 531: To the classes that are being tested, please only add what actually gets tested. If the tests don't check that the content of the docstring is used, then don't add a docstring. Or do test for the docstring.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -468,9 +474,111 @@ def test_add_class_final(parser):
...
+
+        a1: int = 1
```
mauvilsa, line 533: Better to no use the same name in a nested and parent parameters, `a1` here and `a1` in the classes below. If they are different then it it is clearer what comes from where. For example it could be here `i1` (for inner).

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -468,9 +474,111 @@ def test_add_class_final(parser):
...
+
+        Args:
+            a1: a1 help
+        """
+
+        a1: annotated[InnerDataClass, 1] = pydantic.fields.Field(default_factory=InnerDataClass)  # type: ignore[valid-type]
+
+    def test_pydantic_nested_annotated_dataclass(parser: ArgumentParser):
+        parser.add_class_arguments(NestedAnnotatedDataClass, "n")
+        parser.parse_args(["--n.a1.a1=1"])
```
mauvilsa, line 567: Please assert the parsed value. Same in the three tests below.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
             parser.parse_args([f"--model.param={invalid_value}"])
         ctx.match("model.param")
 
+    @pytest.mark.skipif(not pydantic_supports_field_init, reason="Field.init is required")
+    def test_dataclass_field_init_false(self, parser):
+        # This tests the following error:
+        #
+        # TypeError: Parser key "data.p2":
+        # Expected a <class 'str'>. Got value: annotation=str
+        # required=False default='-' init=False
```
mauvilsa, line 727: What is this? How does this test a TypeError? If an error is tested, then I would expect an `pytest.raises` instead of some confusing comment.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
...
+        parser.add_argument("--data", type=PydanticDataFieldInitFalse)
+        cfg = parser.parse_args(["--data.p1=1.0"])
+        assert cfg.data.p1 == 1.0
```
mauvilsa, line 730: Nothing is testing the `init=False`. I would expect here checking that `data.p2` is not in the parsed namespace. And maybe also that `--data.p2` is not in the help output.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
...
+
+    def test_dataclass_stdlib_field(self, parser):
+        parser.add_argument("--data", type=PydanticDataStdlibField)
+        cfg = parser.parse_args(["--data.p1=1.0"])
+        assert cfg.data.p1 == 1.0
```
mauvilsa, line 735: Shouldn't this be testing `data.p2`. And what is the point of having `p1` in the class and tested here?

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
...
+
+    def test_dataclass_stdlib_field_init_with_factory(self, parser):
+        parser.add_argument("--data", type=PydanticDataStdlibFieldWithFactory)
+        cfg = parser.parse_args(["--data.p1=1.0"])
+        assert cfg.data.p1 == 1.0
```
mauvilsa, line 740: Shouldn't this be testing data.p2. And what is the point of having p1 in the class and tested here?

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
...
+
+    def test_dataclass_nested(self, parser):
+        # This tests the following error:
+        #
+        # ValueError: Expected "default" argument to be an instance of
+        # "PydanticData" or its kwargs dict, given
+        # <dataclasses._MISSING_TYPE object at 0x105624c50>
```
mauvilsa, line 747: Same comment as above. What is this? How does this test a ValueError? If an error is tested, then I would expect an `pytest.raises` instead of some confusing comment.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -587,6 +718,37 @@ def test_pydantic_types(self, valid_value, invalid_value, cast, type_str, monkey
...
+
+        parser.add_argument("--data", type=PydanticDataNested)
+        parser.parse_args(["--data", '{"p3": {"p1": 1.0}}'])
```
mauvilsa, line 750: Please assert the parsed values.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -628,3 +798,13 @@ def test_field_factory(self, parser):
         assert cfg1.data.p1 == ["one", "two"]
         assert cfg1.data.p1 == cfg2.data.p1
         assert cfg1.data.p1 is not cfg2.data.p1
+
+    def test_field_init_false(self, parser):
+        # This tests the following error:
+        #
+        # TypeError('Validation failed: Key "data.p2" is required but
+        # not included in config object or its value is None.')
```
mauvilsa, line 806: Here again like the comments above.

jsonargparse_tests/test_dataclass_like.py
```diff
@@ -628,3 +798,13 @@ def test_field_factory(self, parser):
...
+
+        parser.add_argument("--data", type=AttrsFieldInitFalse)
+        cfg = parser.parse_args(["--data.p1=1.0"])
+        assert cfg.data.p1 == 1.0
```
mauvilsa, line 810: Shouldn't this be testing data.p2. And what is the point of having p1 in the class and tested here?

CHANGELOG.rst
```diff
@@ -11,6 +11,14 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
+v4.28.1 (2024-0?-??)
+--------------------
+
+Fixed
+^^^^^
+-  Failure to process ``Annotated`` dataclass members, and inclusion of
+   non-init fields in `attrs` and Pydantic dataclass-like instantiation.
```
mauvilsa, line 20: This pull request is rather big. Maybe it could have been multiple, but it is fine like this. However, it is not a requirement to have only one entry in the change log. The pull request has 6 bullet points, which don't seem all to be reflected here. And I am not sure that it is only fixes. For example `TypeAliasType` was not supported before, so this is a new feature.

Also `TypeAliasType` should be added to the documentation.rst as a supported type.

jsonargparse/_parameter_resolvers.py
```diff
@@ -898,6 +923,23 @@ def get_field_data_attrs(field, name, doc_params):
     )
 
 
+def is_init_field_pydantic_model(field) -> bool:
+    return True
```
mauvilsa, line 927: This is somewhat weird. Why not just a `lambda f: True` where this is used? Then it is clearer there that it is always supported.

mauvilsa (changes requested):
It is looking good. But some work is needed. See the comments.

nkrishnaswami:
> It is looking good. But some work is needed. See the comments.

Thanks for the feedback! I believe I have address these in the latest commit(s).

nkrishnaswami:
That last update broke nested dataclasses on 3.12, and init=False on 3.7 for Pydantic. (The version dependency is a surprise).  I'll take a look at those tonight (UTC-5)

jsonargparse_tests/test_dataclass_like.py
mauvilsa, line 1: One more comment. I saw a warning in python 3.7, saying:
```
PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'init'). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/
```

Certainly it would be preferable to avoid adding support for things that are deprecated. Though, I am not sure what that is about. Or in how many python or OS versions this happens. 

And a related comment. I have made sure that the output of running tests is completely clean. That is, zero warnings. This is so that if something changes, it is immediately obvious there is a problem. If there are tons of warnings then issues are not noticed and identified early. If a test is expected to produce a warning, please capture and assert it. See for example what is done in `test_deprecated.py`.

nkrishnaswami: I had incorrectly believed that the `init` argument to Pydantic `Field` was introduced in 2.4, but it was actually 2.6.  The Python 3.7 coverage test happened to use Pydantic 2.5, so it caught the error (while also producing the warning since the `init` arg had not yet been introduced).

mauvilsa (approved):

PR #494, opened 2024-04-19
[MRG] Support `TypedDict`

### LINKED ISSUE #457: Support of `TypedDict`s
Hi, 
Thanks for this great library! Extremely useful!!

## 🚀 Feature request

The idea here would be to support arguments with type `typing.TypedDict`.

### Motivation

This feature would be great when we want to get a dictionary instead of a namespace but still want to check the keys and types of its elements. 

This could also allow checking the elements in `**kwargs`: https://peps.python.org/pep-0692/

### Pitch

For example, if a function calls another function, we could then pass arguments to the second one very easily:
```python
# script.py
from typing import TypedDict
from jsonargparse import CLI

def aux(a: int = 1):
    print(a)

AuxKwargs = TypedDict("AuxKwargs", aux.__annotation__)

def main(aux_kwargs: AuxKwargs):
    aux(**aux_kwargs)

CLI(main)
```

Then we would be able to call:
```bash
python script.py --aux_kwargs.a=2
```

### Alternatives

Currently, the only alternative for such cases is to use the `dict` type, which prevents from checking the key names and types individually.
In the above example, the current alternative is:
```python
def main(aux_kwargs: dict):
    aux(**aux_kwargs)
```

PierreGtch:
Hi,
Thanks for this great library!

## What does this PR do?

This PR adds support to `TypedDict`s.
Closes #457

## Before submitting

I updated the `adapt_typehints` function, added tests in `test_typehints.py`, and updated the `CHANGELOG.rst`. Please tell me if I missed anything :)

### DIFF
--- a/jsonargparse_tests/test_typehints.py
+++ b/jsonargparse_tests/test_typehints.py
@@ -20,9 +20,11 @@ from typing import (
     Set,
     Tuple,
     Type,
-    TypedDict,
     Union,
 )
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
 from unittest import mock
 from warnings import catch_warnings
 
@@ -491,6 +493,8 @@ def test_mapping_nested_without_args(parser):
 
 
 def test_typeddict_without_arg(parser):
+    if sys.version_info < (3, 8):
+        pytest.skip("TypedDict introduced in python 3.8")
     parser.add_argument("--typeddict", type=TypedDict("MyDict", {}))
     assert {} == parser.parse_args(["--typeddict={}"])["typeddict"]
     pytest.raises(ArgumentError, lambda: parser.parse_args(['--typeddict={"a":1}']))
@@ -498,6 +502,8 @@ def test_typeddict_without_arg(parser):
 
 
 def test_typeddict_with_args(parser):
+    if sys.version_info < (3, 8):
+        pytest.skip("TypedDict introduced in python 3.8")
     parser.add_argument("--typeddict", type=TypedDict("MyDict", {"a": int}))
     assert {"a": 1} == parser.parse_args(["--typeddict={'a': 1}"])["typeddict"]
     # extra key
@@ -510,6 +516,8 @@ def test_typeddict_with_args(parser):
 
 
 def test_typeddict_with_args_ntotal(parser):
+    if sys.version_info < (3, 8):
+        pytest.skip("TypedDict introduced in python 3.8")
     parser.add_argument("--typeddict", type=TypedDict("MyDict", {"a": int}, total=False))
     assert {"a": 1} == parser.parse_args(["--typeddict={'a': 1}"])["typeddict"]
     assert {} == parser.parse_args(["--typeddict={}"])["typeddict"]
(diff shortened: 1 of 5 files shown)


PierreGtch:
Ready for review :)

jsonargparse_tests/test_typehints.py
```diff
@@ -489,6 +492,42 @@ def test_mapping_nested_without_args(parser):
     assert {"b": {"c": 2}} == parser.parse_args(['--map={"b": {"c": 2}}']).map
 
 
+def test_typeddict_without_arg(parser):
+    if sys.version_info < (3, 8):
+        pytest.skip("TypedDict introduced in python 3.8")
```
mauvilsa, line 497: Be consistent with the rest of the code. In other tests, skips are done using `@pytest.mark.skipif`, so please do the same in the added tests. See for example: https://github.com/omni-us/jsonargparse/blob/af37f088f257a184f1f9e77812539d8a2e584276/jsonargparse_tests/test_typehints.py#L525

jsonargparse_tests/test_typehints.py
```diff
@@ -489,6 +492,42 @@ def test_mapping_nested_without_args(parser):
...
+    parser.add_argument("--typeddict", type=TypedDict("MyDict", {}))
+    assert {} == parser.parse_args(["--typeddict={}"])["typeddict"]
+    pytest.raises(ArgumentError, lambda: parser.parse_args(['--typeddict={"a":1}']))
```
mauvilsa, line 500: Better if you assert part of the exception messages. For example in this case, it could be asserted that the error says `Unexpected keys`. For how to do it, see for example https://github.com/omni-us/jsonargparse/blob/af37f088f257a184f1f9e77812539d8a2e584276/jsonargparse_tests/test_typehints.py#L54-L56

jsonargparse_tests/test_typehints.py
```diff
@@ -489,6 +492,42 @@ def test_mapping_nested_without_args(parser):
...
+    pytest.raises(ArgumentError, lambda: parser.parse_args(["--typeddict=1"]))
+
+
+def test_typeddict_with_args(parser):
+    if sys.version_info < (3, 8):
+        pytest.skip("TypedDict introduced in python 3.8")
+    parser.add_argument("--typeddict", type=TypedDict("MyDict", {"a": int}))
+    assert {"a": 1} == parser.parse_args(["--typeddict={'a': 1}"])["typeddict"]
+    # extra key
```
mauvilsa, line 509: Instead of comments, use `subtests` like done for example here
https://github.com/omni-us/jsonargparse/blob/af37f088f257a184f1f9e77812539d8a2e584276/jsonargparse_tests/test_typehints.py#L664

mauvilsa (changes requested):
Thank you for contributing!

The changes mostly look good. One thing that is missing is a change in `DOCUMENTATION.rst`. For this feature, maybe just extend the following bullet point mentioning the support for `TypedDict`.
https://github.com/omni-us/jsonargparse/blob/af37f088f257a184f1f9e77812539d8a2e584276/DOCUMENTATION.rst?plain=1#L403-L404

PierreGtch:
All done @mauvilsa :)

mauvilsa (approved):

Issue #301, opened 2023-06-12
Question: is it possible to make CLI replace parameter dashes with underscores?


iskandr:
I'm curious about replacing argparse with jsonargparse in some of my projects since simply running `CLI(main_fn)` would cover my most common use cases. However, it's not clear to me if there's any hook to change the mapping from Python argument names to CLI parameters. Specifically, I typically have multi-word parameters represented with dashes as commandline arguments. So, `arg_like_this` in Python comes `--arg-like-this` on the commandline. Is there any way to replicate this behavior with jsonargparse?

Thanks!

mauvilsa:
>  `arg_like_this` in Python comes `--arg-like-this` on the commandline.

Currently no. But could be added.

iskandr:
@mauvilsa I don't mind digging into the code if you have a hunch about the most harmonious direction to try. Should I add it as a specific option to `CLI` (eg `convert_underscores_to_dashes=True`) or make a more general transform function (eg `commandline_name_transform=lambda x: x.replace("_", "-")`)?

iskandr:
Thanks for your reply @mauvilsa. 

I can start with the motivation and circle back on implementation details once I get more familiar with this repo. 

I would want hyphens/dashes (instead of underscores) to maintain consistency with other tools I have developed over the past decade. It would be very confusing to have to switch between dashes and underscores depending on the tool. 

More generally, dashes are the GNU standard (https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) and (somewhat subjectively) seem to be preferred by many other library/tool maintainers (eg https://github.com/pytorch/pytorch/issues/65052, https://github.com/rspec/rspec-core/issues/497).

mauvilsa:
Some thoughts:

- Given that the motivation is only to allow more standard **command line** behavior (hyphens), it does make sense that for config files, the original parameter names are used (underscores).
- The `--help` is not just to explain individual command line options, but also config files and environment variables. Thus, the help would need to show both with hyphens and underscores.
- Having both underscores and hyphens is required in cases where there is a need for backward compatibility, e.g. [pytorch#65052](https://github.com/pytorch/pytorch/issues/65052).
- Implementation-wise it is not possible to have an argument that uses one name for command and another for the rest. Thus, the feature seems more viable as "add hyphens alias" instead of "replacing underscore with hyphens".
[...]

PR #532, opened 2024-06-19
Initial scriptconfig support

Erotemic:
## What does this PR do?

This PR is an initial attempt at https://github.com/omni-us/jsonargparse/issues/244 

This allows a user to define a scriptconfig object that gives them fine-grained control over the arguments and metadata on the CLI, without requiring the variables to be typed more than once. In other words, the signature of a function is maintained in a class, and it is assumed that the keyword arguments given to that class will be used to create an instance of the scriptconfig object.

Here is what the MWE looks like:

```python
import scriptconfig as scfg


class MyClassConfig(scfg.DataConfig):
    key1 = scfg.Value(1, alias=['key_one'], help='description1')
    key2 = scfg.Value(None, type=str, help='description1')
    key3 = scfg.Value(False, isflag=True, help='description1')
    key4 = 123
    key5 = '123'


class MyClass:
    __scriptconfig__ = MyClassConfig

    def __init__(self, regular_param1, **kwargs):
        self.regular_param1 = regular_param1
        self.config = MyClassConfig(**kwargs)


def main():
    import jsonargparse
    import shlex
    import ubelt as ub
    import rich
    from rich.markup import escape
    parser.add_class_arguments(MyClass, nested_key='my_class', fail_untyped=False, sub_configs=True)
    parser.add_argument('--foo', default='bar')
    parser.add_argument('-b', '--baz', '--buzz', default='bar')
    print('Parse Args')
    cases = [
        '',
        '--my_class.key1 123',
        '--my_class.key_one 123ab',
        '--my_class.key4 strings-are-ok',
    ]
    for case_idx, case in enumerate(cases):
        print('--- Case {case_idx} ---')
        print(f'case={case}')
        args = shlex.split(case)
        config = parser.parse_args(args)
        instances = parser.instantiate_classes(config)

        my_class = instances.my_class
        rich.print(f'config = {escape(ub.urepr(config, nl=2))}')
        rich.print(f'my_class.config = {escape(ub.urepr(my_class.config, nl=2))}')
        print('---')


if __name__ == '__main__':
    """
    CommandLine:
        cd ~/code/geowatch/dev/mwe/
        python jsonargparse_scriptconfig_integration_test.py
    """
    main()
```

You'll note that it looks similar to dataclass / pydantic / attrs object. However,  a major difference is that it doesn't rely on type annotations to store metadata. It uses a special "Value" class, which can be augmented with things like help, aliases, and other information useful to building both a CLI and a function / class signature.

The main scriptconfig page is here for more information: https://gitlab.kitware.com/utils/scriptconfig

I've been using a monkey-patched version of jsonargparse that allows me to work with scriptconfig objects for over a year now, and in late 2023 there was some change to jsonargparse that broke me. This force me to pin jsonargparse and pytorch_lightning to older versions, and I've finally had time to look into it. However, I'd really really really like is there was something codified into jsonargparse that would either directly support scriptconfig or expose some way for users to do custom things when running `add_class_arguments` based on a property in the class itself (which will let it integrate with lightning much easier). I don't particularly care if scriptconfig itself is supported, what I need is something that won't get deprecated or refactored that I can hook into.

So far this PR just adds basic support for scriptconfig itself. This involves some extra logic in `_add_scriptconfig_arguments`, and in order to support the alias feature of scriptconfig, I updated `ParamData` so it is now equipped with a `short_aliases` and a `long_aliases` attribute. I've also added a method to it so it can construct the args that will ultimately be passed to argparse. This makes modifications to `_add_signature_parameter` a bit cleaner, and also sets the stage to allow other argument sources to specify aliases.

## Before submitting

- [ ] Did you read the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)?
- [ ] Did you verify that new and existing **tests pass locally**?

WRT to this checklist, I don't want to invest too much time into this before having a discussion with the authors.

I'm looking for feedback on the level of support the maintainers would be willing to provide for something like this. I think the simple thing to do would be to "add scriptconfig support" and be done with it, in which case I could clean up the existing code. The alternative is to allow allow classes to have some attribute (e.g. `__customize_signature__`) that lets the user return a list of additional `ParamData` objects that should be recognized by the CLI. However, that involves making `ParamData` a public class, so that has its own set of tradeoffs.

### DIFF
--- a/jsonargparse/_parameter_resolvers.py
+++ b/jsonargparse/_parameter_resolvers.py
@@ -44,6 +44,22 @@ class ParamData:
     component: Optional[Union[Callable, Type, Tuple]] = None
     parent: Optional[Union[Type, Tuple]] = None
     origin: Optional[Union[str, Tuple]] = None
+    short_aliases: Optional[List[str]] = None
+    long_aliases: Optional[List[str]] = None
+
+    def _resolve_args_and_dest(self, is_required=False, as_positional=False, nested_key: Optional[str] = None):
+        name = self.name
+        dest = (nested_key + "." if nested_key else "") + name
+        if is_required and as_positional:
+            args = [dest]
+        else:
+            long_names = [name] + list((self.long_aliases or []))
+            short_names = list(self.short_aliases or [])
+            nest_prefix = nested_key + "." if nested_key else ""
+            short_option_strings = ["-" + nest_prefix + n for n in short_names]
+            long_option_strings = ["--" + nest_prefix + n for n in long_names]
+            args = short_option_strings + long_option_strings
+        return args, dest
 
 
 ParamList = List[ParamData]
--- a/jsonargparse/_signatures.py
+++ b/jsonargparse/_signatures.py
@@ -285,6 +285,7 @@ class SignatureArguments(LoggerProperty):
 
         ## Add parameter arguments ##
         added_args: List[str] = []
+
         for param in params:
             self._add_signature_parameter(
                 container,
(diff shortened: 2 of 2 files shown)


mauvilsa:
@Erotemic thank you for the pull request. Scriptconfig support like proposed here most likely won't happen. But I will have this in mind to add some public interface that would allow you to do this. Note that this is not a simple topic. New features should behave well with all other existing features and not prevent or make more difficult other potential new features. For example what is done here for aliases could conflict with #301. Please just be patient since this will take time.

For the time being we can use this pull request to test that your integration works with the latest changes.

Erotemic:
Thanks for taking the time to look at / review this. Avoiding integration with a particular library makes a lot of sense, but I'm glad you're open to some public API to make something like this possible.

My thought is that `ParamData` is the obvious thing to expose to the user. If there was something like:

```python
class MyClass:
    __jsonargparse_hook__ = ...

import jsonargparse
parser.add_class_arguments(MyClass, nested_key='my_class', fail_untyped=False, sub_configs=True)
```

Where `__jsonargparse_hook__` (or a better name) was a callback that users could use to return a custom list of `ParamData` objects, then that would allow me to write something specific for scriptconfig to integrate with jsonargparse without jsonargparse having direct knowledge of specific consumer packages. Perhaps the signature looks like:

```python
def __jsonargparse_hook__(params: List[ParamData]) -> List[ParamData]:
    """
    Input is the initial list of default parameters that would be added. Modify / add / remove these 
    and return the modified list to control details of which arguments are exposed to jsonargparse. 
    """
```

However, `ParamData` may need to be modified (or given an initial restricted public form that is added to as the default develops). I think the first step is to add support for aliases. It looks like I made a PR to add alias last year that I forgot about: https://github.com/omni-us/jsonargparse/pull/255 but it is outdated and it looks like things have changed.

I don't have as much time to work on FOSS code in the next few months, but jsonargparse is a critical enough part of my workflow that I can prioritize it, but I want to have a solid plan before I commit to anything. 

Let me know what you think.

mauvilsa:
Thank you for the proposal. Though, I don't like it much. I see several issues with it.

- Seems to mostly consider the scriptconfig integration as implemented in this pull request, instead of targeting a more general scope. If a public API is added, it must be clear that several cases could benefit from it, instead of just a single library.
- Only works for classes, leaving out functions and methods.
- Not clear how this works when implementing subclasses. 
- It goes against a principle that jsonargparse has been following: have a design that promotes some good practices. Among these are separation of concerns and low coupling. For example note how [cli.py](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/cli.py) is highly decoupled from the rest of `lightning` or user implemented module classes. The idea is, it is better that all the logic of a CLI be implemented in a single place, and not requiring a bunch of changes throughout a code base. That is, I don't like this `__jsonargparse_hook__` in classes. I know that scriptconfig works like this. But that is no reason to do it similar in jsonargparse.

In my previous comment I was asking for patience because right now I am not sure how the integration API would be. And it will take some time for this, since I am focused on other stuff right now. But I will propose something here when ready.

Erotemic:
Closing as this is not planned to be integrated, although I will continue to maintain this branch as my use-cases require some sort of way to define parameters without relying on the python annotation system or docstrings.

PR #573, opened 2024-09-12
Fix tests inside of docker

Erotemic:
I've run into a tricky issue when running tests locally inside of a docker environment. Tests that involved changing permissions to make a file unreadable were failing when I ran them inside of docker, but they worked on my host system. After a bit of thinking I realized this was because the default user in docker is root, which means files can be read even if they are unreadable according to the permissions.

To address this, I made two changes:
* In test_core, I check if I can still read the file after it is forced to be unreadable. If it can be, then we can assume we are in some special circumstance, and we should skip the test.

* In test_util, it looks like there is a custom path object, so I fell back to just checking if `os.geteuid() == 0`, which should only be true on linux for the root user. In this case I skip the test.

Because this is only a test change, I'm not sure if you want a changelog entry.

### DIFF
--- a/jsonargparse_tests/test_core.py
+++ b/jsonargparse_tests/test_core.py
@@ -39,6 +39,7 @@ from jsonargparse_tests.conftest import (
     skip_if_fsspec_unavailable,
     skip_if_not_posix,
     skip_if_responses_unavailable,
+    skip_if_running_as_root,
 )
 
 
@@ -329,6 +330,7 @@ def test_parse_path_defaults(parser, tmp_cwd):
 
 
 @skip_if_not_posix
+@skip_if_running_as_root
 def test_parse_path_file_not_readable(parser, tmp_cwd):
     config_path = Path("config.json")
     config_path.touch()
@@ -792,6 +794,7 @@ def test_default_config_file_invalid_value(parser, tmp_cwd):
 
 
 @skip_if_not_posix
+@skip_if_running_as_root
 def test_default_config_file_unreadable(parser, tmp_cwd):
     default_config_file = Path("defaults.yaml")
     default_config_file.write_text("op1: from yaml\n")
--- a/jsonargparse_tests/test_util.py
+++ b/jsonargparse_tests/test_util.py
@@ -47,6 +47,7 @@ from jsonargparse_tests.conftest import (
     skip_if_fsspec_unavailable,
     skip_if_requests_unavailable,
     skip_if_responses_unavailable,
+    skip_if_running_as_root,
 )
 
 if responses_available:
@@ -136,6 +137,7 @@ def test_path_equality_operator(paths):
     assert Path("123", "fc") != 123
 
 
+@skip_if_running_as_root
 def test_path_file_access_mode(paths):
     Path(paths.file_rw, "frw")
     Path(paths.file_r, "fr")
@@ -150,6 +152,7 @@ def test_path_file_access_mode(paths):
     pytest.raises(TypeError, lambda: Path("file_ne", "fr"))
 
 
+@skip_if_running_as_root
 def test_path_dir_access_mode(paths):
     Path(paths.dir_rwx, "drwx")
     Path(paths.dir_rx, "drx")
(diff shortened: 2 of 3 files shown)


mauvilsa:
Running tests as root doesn't sound like a great idea. Better to use a real user so that the tests are run in an environment closer to how the code will be used later. And if you want to run the tests, skipping tests shouldn't be the go to solution.

I would recommend that you run the tests as a non-root user. In docker it is easy to switch user. If for some reason you really don't want to do that, from the command line you can tell pytest to skip some tests. No need to change the code.
```
pytest -k "not <test-name-1> and not <test-name-2> ..."
```

Erotemic:
Yes, it is better to use a real use. Unfortunately, running as root in docker is the de-facto standard. Do I like it? No. But it is the reality.

I feel that just running `pytest` in the repo should "just work" in reasonable circumstances. But I also feel I'm going to have no success in arguing a point here. :face_exhaling:

mauvilsa:
I have no problem in accepting this pull request. If you really want to go ahead a run tests as root, I can comment a couple of things and later we merge it.

But a clarification. Running tests as root is not a standard. The tests run here via circle ci all use docker images. They run as non-root and I didn't anything special for this. Maybe the github tests also run in docker containers, I am not sure, but if they do, then they run as non-root by default.

Base docker images do come as default as root. Though, the idea is that layers are added on top and the user changed. It is a good practice to never run docker containers as root.

jsonargparse_tests/test_util.py
```diff
@@ -137,6 +137,8 @@ def test_path_equality_operator(paths):
 
 
 def test_path_file_access_mode(paths):
+    if is_posix and os.geteuid() == 0:
+        pytest.skip("User is root, skipping permission tests")
```
mauvilsa, line 141: Please skip using a decorator, similar to https://github.com/omni-us/jsonargparse/blob/1bab38ae7a5fe281c62fe784fc620bbbf259f2e8/jsonargparse_tests/conftest.py#L35-L38

jsonargparse_tests/test_core.py
```diff
@@ -801,7 +806,15 @@ def test_default_config_file_unreadable(parser, tmp_cwd):
...
-    assert parser.get_default("op1") == "from default"
+    try:
+        # Check if the file is actually unreadable.
+        default_config_file.read_text()
+    except PermissionError:
+        assert parser.get_default("op1") == "from default"
+    else:
+        # Sometimes chmod 0 will not make the file unreadable
+        # (e.g. if running as root in a docker container).
+        pytest.skip("File was still readable even after chmod(0)")
```
mauvilsa, line 817: Please skip the entire test function using a decorator, instead of this try-except block.

jsonargparse_tests/test_core.py
```diff
@@ -333,7 +333,12 @@ def test_parse_path_file_not_readable(parser, tmp_cwd):
     config_path = Path("config.json")
     config_path.touch()
     config_path.chmod(0)
-    pytest.raises(TypeError, lambda: parser.parse_path(config_path))
+    try:
+        config_path.read_text()
+    except PermissionError:
+        pytest.raises(TypeError, lambda: parser.parse_path(config_path))
+    else:
+        pytest.skip("File was still readable after chmod(0)")
```
mauvilsa, line 341: Please skip the entire test function using a decorator, instead of this try-except block.

mauvilsa (changes requested):

Erotemic:
> If you really want to go ahead a run tests as root, I can comment a couple of things and later we merge it.

Thanks. I wish running as non-root was an option, but I'm working in an environment where everything is running on a base `python:3.11` image and there is little-to-no opportunity to change that. Others will install their packages to hard coded paths (e.g. `/root/<theirstuff>`) so if I swapped out the shared base image for something with a builtin non-root user then that would cause breakage. For better or worse I'm stuck in this environment, and this isn't the first time I've worked in a team where root-user docker images was the norm. For my part, I just want to be able to run through test suites (and show others how to run through the tests suites) with an invocation of `pytest`.

Having a root-user environment certainly isn't a problem for professional CI systems, but when working on teams with developers at many different skill levels, it can and does come up. I appreciate your willingness to help me out by adding code to detect and work around this admittedly non-ideal enviornment.

mauvilsa (approved):

Issue #590, opened 2024-10-05
Selecting a subtype raises an exception in Python 3.12.7


eggry:
### To reproduce
First, create a file named `demo.py` with the following content:
```python
from typing import Optional
from jsonargparse import CLI

class Foo: ...

class Bar(Foo):
    def __init__(self, val: int) -> None:
        self.val = val

def main(data: Optional[Foo] = None) -> None:
    print(data)

if __name__ == "__main__":
    CLI(main)
```
Then, run the following command in your terminal:
```bash
python demo.py --data Bar --data.val 1
```
The output is:
```
usage: demo.py [-h] [--config CONFIG] [--print_config[=flags]] [--data.help CLASS_PATH_OR_NAME] [--data DATA]
error: cannot unpack non-iterable ActionTypeHint object
```

Such behavior will also affect LitGPT. For example, the command
```bash
    litgpt finetune tiiuae/falcon-7b \
        --data JSON \
        --data.json_path data/mydata.json \
        --out_dir data/mydata-finetuned
```
will also result in the same error because `JSON` is a subclass of `DataModule`.

### Expected behavior
On Python 3.12.6, the output is: 
```
<__main__.Bar object at 0x*******>
```

### Environment

- jsonargparse version: 4.32.1
- Python version: 3.12.7
- How jsonargparse was installed: `pip install jsonargparse==4.32.1`
- OS: 22.04.3-Ubuntu

mauvilsa:
Fixed with #591

PR #591, opened 2024-10-05
Fix issue #590: Behavior change in `argparse._parse_optional` breaks the parse of optional arguments

eggry:
## What does this PR do?

In Python 3.13 and 3.12.7, the behavior of `_parse_optional` has changed to return a list of tuples: https://github.com/python/cpython/pull/124631
This change would break the argument parser of this library.
For example, the existing unit test failed, resulting in:
```
FAILED jsonargparse_tests/test_dataclass_like.py::test_optional_dataclass_single_param_change - argparse.ArgumentError: cannot unpack non-iterable ActionTypeHint object
``` 

This PR will detect this behavior and return the proper format accordingly.
This PR resolves issue #590.

## Before submitting

- [n/a] Did you write **unit tests** such that there is 100% coverage on related code? (required for bug fixes and new features)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,14 @@ The semantic versioning only considers the public API as described in
 paths are considered internals and can change in minor and patch releases.
 
 
+v4.33.2 (2024-10-??)
+--------------------
+
+Fixed
+^^^^^
+- Fix `parse_argv_item` to address the behavior change in `argparse` since Python 3.12.7 (`#591 <https://github.com/omni-us/jsonargparse/pull/591>`__).
+
+
 v4.33.1 (2024-09-26)
 --------------------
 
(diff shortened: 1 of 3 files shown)


mauvilsa (changes requested):
Thank you very much for the fix!

Please add at the top of the changelog an entry for this fix, in a new section, like:
```
v4.33.2 (2024-10-??)
--------------------

Fixed
^^^^^
- ...
```

Also, since this breaking change also affects [argcomplete](https://github.com/kislyuk/argcomplete/issues/507) and we don't know when there will be a new release with the fix, please add to `test_argcomplete.py` the following:
```python
@pytest.fixture(autouse=True)
def temporal_skip_if_py312():
    if sys.version_info[:2] == (3, 12):
        pytest.skip("currently failing on 3.12 due to argparse breaking change")
```

mauvilsa (approved):

PR #592, opened 2024-10-06
fix: empty tuples are correctly parsed

### LINKED ISSUE #589: Empty tuple default value not supported
## 🐛 Bug report

Empty tuples are not supported by the parser.

### To reproduce

```

@dataclass
class TestCfg:
    a: tuple[int, ...] = ()


if __name__ == "__main__":
    parser = ArgumentParser(description="CLI for training models", exit_on_error=False)
    parser.add_argument("test", type=TestCfg)
    args = parser.parse_args()
    print(args)
```
Running the following program without passing any arguments or by passing an empty list/tuple `"[]"` results in the following error:
```
argparse.ArgumentError: Validation failed: Parser key "test.a":
  Expected a non-empty tuple. Got value: []
  ```

### Expected behavior

To be able to parse empty tuples without errors.

### Environment

- jsonargparse version: 4.33.1
- Python version: 3.10
- OS: MacOS

### PR DESCRIPTION
**Pull request by MiguelMonteiro**
## What does this PR do?

Empty tuples are now successfully parsed.

Fixes #589

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,7 @@ The semantic versioning only considers the public API as described in
 paths are considered internals and can change in minor and patch releases.
 
 
-v4.33.3 (2024-10-09)
+v4.33.3 (2024-10-??)
 --------------------
 
 Fixed
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,16 @@ The semantic versioning only considers the public API as described in
 paths are considered internals and can change in minor and patch releases.
 
 
+v4.33.2 (2024-10-09)
+--------------------
+
+Fixed
+^^^^^
+- Empty tuples are now parsed correctly instead of raising an error.
+  (`#592 <https://github.com/omni-us/jsonargparse/pull/592`__).
+
+
+
 v4.33.1 (2024-09-26)
 --------------------
 
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -847,8 +847,6 @@ def adapt_typehints(
             for n, v in enumerate(val):
                 subtypehint = subtypehints[0 if is_ellipsis or not is_tuple else n]
                 val[n] = adapt_typehints(v, subtypehint, **adapt_kwargs)
-            if is_tuple and len(val) == 0:
-                val = ()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: jsonargparse/_typehints.py [hunk 1]
```diff
@@ -848,7 +848,7 @@ def adapt_typehints(
                 subtypehint = subtypehints[0 if is_ellipsis or not is_tuple else n]
                 val[n] = adapt_typehints(v, subtypehint, **adapt_kwargs)
             if is_tuple and len(val) == 0:
-                raise_unexpected_value("Expected a non-empty tuple", val)
```
mauvilsa, line 851: Don't remove the raise for empty tuple. This is correct and necessary. Only tuples with `...` should be allowed to be empty. But other tuples not. The code here should be:
```python
if is_tuple and len(val) == 0 and not is_ellipsis:
    raise_unexpected_value("Expected a non-empty tuple", val)
```

mauvilsa (changes requested):

jsonargparse/_typehints.py
mauvilsa: Actually no. These two lines should be removed. Number of elements in the tuple are already handled by
```python
if is_tuple and not is_ellipsis and len(val) != len(subtypehints):
    raise_unexpected_value(f"Expected a tuple with {len(subtypehints)} elements", val)
```
And converting the empty list to a tuple would already be handled by
```python
if not serialize:
    val = tuple(val) if typehint_origin in {Tuple, tuple} else set(val)
```

mauvilsa (changes requested):
Also please add a changelog entry in a new section as explained in the contributions guide.

MiguelMonteiro:
Thank you for the feedback, I have made the requested changes, let me know if it's correct.

CHANGELOG.rst
```diff
@@ -12,6 +12,15 @@ The semantic versioning only considers the public API as described in
 paths are considered internals and can change in minor and patch releases.
 
 
+v4.33.3 (2024-10-09)
```
mauvilsa, line 15:
```suggestion
v4.33.3 (2024-10-??)
```

mauvilsa (changes requested):

PR #599, opened 2024-10-15
Fix/callable protocols 2

### LINKED ISSUE #595: Callable protocols inheritance broken
## 🐛 Bug report

Callable protocols (implement `__call__`) seems to be parser incorrectly in comparison to "normal" protocols.

### To reproduce

Example that works
```
from typing import Protocol

class BaseGood(Protocol):
    def predict(self, a: int) -> int: ...


@dataclass
class SpecializedGood(BaseGood):
    number: int = 5

    def predict(self, a: int) -> int:
        return a + 5


def main() -> None:
    parser = ArgumentParser(description="CLI for training models", exit_on_error=False)
    parser.add_argument("--cfg", type=BaseGood)
    args = parser.parse_args(["--cfg=SpecializedGood", "--cfg.number=5"])
    print(args)
    
if __name__ == "__main__":
    main()
```

Example that does not work:
```
class BaseBad(Protocol):
    def __call__(self, a: int) -> int: ...


@dataclass
class SpecializedBad(BaseBad):
    number: int = 5

    def __call__(self, a: int) -> int:
        return a + 5


def main() -> None:
    parser = ArgumentParser(description="CLI for training models", exit_on_error=False)
    parser.add_argument("--cfg", type=BaseBad)
    args = parser.parse_args(["--cfg=SpecializedBad", "--cfg.number=5"])
    print(args)
    
if __name__ == "__main__":
    main()
```

The second example raises the following error:
```
Import path SpecializedBad does not correspond to a subclass of BaseBad
```
Seems like the parser does not match the inherited protocol signature to the base if the protocol methods are private (?).

### Expected behavior

I would expect the second example to work like the first.

### Environment

- jsonargparse version: 4.33.1
- Python version: 3.10
- OS: MacOS

MiguelMonteiro:
## What does this PR do?

Fixes callable protocol inheritance by allowing a list of private methods to be considered when matching signatures for inheritance of subclasses.  The set of `allowed_dunder_methods` currently only contains the `__call__` method.

Fixes #595

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,13 +11,6 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
-v4.34.1 (2024-10-??)
---------------------
-
-Fixed
-^^^^^
-- Fix callable protocol inheritance.
-
 
 v4.34.0 (2024-10-??)
 --------------------
@@ -40,6 +33,8 @@ Fixed
   <https://github.com/omni-us/jsonargparse/pull/608>`__).
 - Failure when resolving forward references from dataclass parameter types
   (`#611 <https://github.com/omni-us/jsonargparse/pull/611>`__).
+- Fix callable protocol inheritance.
+  (`#599 <https://github.com/omni-us/jsonargparse/pull/599>`__).
 
 Changed
 ^^^^^^^
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,13 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
+v4.34.1 (2024-10-??)
+--------------------
+
+Fixed
+^^^^^
+- Fix callable protocol inheritance.
+
 
 v4.34.0 (2024-10-??)
 --------------------
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -1101,6 +1101,7 @@ def adapt_typehints(
 
 
 def implements_protocol(value, protocol) -> bool:
+    allowed_dunder_methods = {"__call__"}
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
@@ -1108,7 +1109,7 @@ def implements_protocol(value, protocol) -> bool:
         return False
     members = 0
     for name, _ in inspect.getmembers(protocol, predicate=inspect.isfunction):
-        if name.startswith("_"):
+        if name.startswith("_") and name not in allowed_dunder_methods::
             continue
         if not hasattr(value, name):
             return False
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -1087,8 +1087,23 @@ def adapt_typehints(
     return val
 
 
+protocol_irrelevant_dunder_methods = {
+    "__init__",
+    "__new__",
+    "__del__",
+    "__getattr__",
+    "__getattribute__",
+    "__setattr__",
+    "__delattr__",
+    "__reduce__",
+    "__reduce_ex__",
+    "__getstate__",
+    "__setstate__",
+    "__subclasshook__",
+}
+
+
 def implements_protocol(value, protocol) -> bool:
-    allowed_dunder_methods = {"__call__"}
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
@@ -1096,7 +1111,8 @@ def implements_protocol(value, protocol) -> bool:
         return False
     members = 0
     for name, _ in inspect.getmembers(protocol, predicate=inspect.isfunction):
-        if name.startswith("_") and name not in allowed_dunder_methods:
+        is_dunder = name.startswith("__") and name.endswith("__")
+        if (not is_dunder and name.startswith("_")) or (is_dunder and name in protocol_irrelevant_dunder_methods):
             continue
         if not hasattr(value, name):
             return False
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -1105,7 +1105,7 @@ def implements_protocol(value, protocol) -> bool:
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
-    if not inspect.isclass(value):
+    if not inspect.isclass(value): # Should we check if callables implement protocols?
         return False
     members = 0
     for name, _ in inspect.getmembers(protocol, predicate=inspect.isfunction):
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -11,6 +11,13 @@ The semantic versioning only considers the public API as described in
 :ref:`api-ref`. Components not mentioned in :ref:`api-ref` or different import
 paths are considered internals and can change in minor and patch releases.
 
+v4.34.1 (2024-10-??)
+--------------------
+
+Fixed
+^^^^^
+- Fix callable protocol inheritance.
```
mauvilsa, line 19: v4.34.0 hasn't been released yet, so this entry should be there instead of a new v4.34.1 section. Also please add a link to this pull request.

jsonargparse/_typehints.py
```diff
@@ -1101,14 +1101,15 @@ def adapt_typehints(
...
+    allowed_dunder_methods = {"__call__"}
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
     if not inspect.isclass(value):
         return False
     members = 0
     for name, _ in inspect.getmembers(protocol, predicate=inspect.isfunction):
-        if name.startswith("_"):
+        if name.startswith("_") and name not in allowed_dunder_methods:
```
mauvilsa, line 1112: Give me some time and I analyze the dunder methods in general and maybe in this pull request we do more than `__call__`.

mauvilsa (changes requested):
Please add a unit test for a protocol with `__call__`.

MiguelMonteiro:
> Please add a unit test for a protocol with `__call__`.

I added some tests for callable protocols. One question that came up is whether functions should be checked against the protocol. Currently that test is failing, here is the example:
```
class CallableInterface(Protocol):
    def __call__(self, items: List[float]) -> List[float]: ...


class ImplementsCallableInterface1:
    def __init__(self, batch_size: int):
        self.batch_size = batch_size

    def __call__(self, items: List[float]) -> List[float]:
        return items

def implements_callable_interface2(items: List[float]) -> List[float]:
    return items
 ```
 The result is:
 ```
 implements_protocol(ImplementsCallableInterface1, CallableInterface) = True
implements_protocol(implements_callable_interface2, CallableInterface) = False
 ```

File: jsonargparse/_typehints.py [hunk 3]
```diff
@@ -1101,14 +1101,15 @@ def adapt_typehints(
...
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
-    if not inspect.isclass(value):
+    if not inspect.isclass(value):  # Should we check if callables implement protocols?
         return False
     members = 0
     for name, _ in inspect.getmembers(protocol, predicate=inspect.isfunction):
-        if name.startswith("_"):
+        if name.startswith("_") and name not in allowed_dunder_methods:
```
mauvilsa, line 1112:
```suggestion
        is_dunder = name.startswith("__") and name.endswith("__")
        if (not is_dunder and name.startswith("_")) or (is_dunder and name in protocol_irrelevant_dunder_methods):
```

This is what I came up with.

jsonargparse/_typehints.py
```diff
@@ -1101,14 +1101,15 @@ def adapt_typehints(
 
 
 def implements_protocol(value, protocol) -> bool:
+    allowed_dunder_methods = {"__call__"}
```
mauvilsa, line 1104:
```suggestion
protocol_irrelevant_dunder_methods = {
    "__init__", "__new__", "__del__", "__getattr__", "__getattribute__", "__setattr__",
    "__delattr__", "__reduce__", "__reduce_ex__", "__getstate__", "__setstate__"
}

def implements_protocol(value, protocol) -> bool:
```

jsonargparse/_typehints.py
```diff
@@ -1101,14 +1101,15 @@ def adapt_typehints(
...
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
-    if not inspect.isclass(value):
+    if not inspect.isclass(value):  # Should we check if callables implement protocols?
```
mauvilsa, line 1108: It does make sense that a function could implement a callable protocol. But I would leave this for a different pull request.

mauvilsa (changes requested):

MiguelMonteiro:
Thanks for the feedback, sorry it's been really busy at work, I will try to get to this next week

MiguelMonteiro:
I made the necessary fixes, currently only one subtest is failing but I don't fully understand why.

```
@pytest.mark.parametrize(
    "expected, value",
    [
        (True, ImplementsCallableInterface1),
        (False, ImplementsCallableInterface1(1)),
        (
            False,
            implements_callable_interface2,
        ),  # TODO: switch to True once functions that implement protocol are checked correctly
        (False, NotImplementsCallableInterface1),
        (False, NotImplementsCallableInterface2),
        (False, NotImplementsCallableInterface3),
        (False, not_implements_callable_interface4),
        (False, object),
    ],
)
def test_implements_callable_protocol(expected, value):
    assert implements_protocol(value, CallableInterface) is expected
```

The last subtest ` (False, object)` raises the following error instead of returning `False`

```FAILED jsonargparse_tests/test_subclasses.py::test_implements_callable_protocol[False-object] - ValueError: Invalid or unsupported input: class=<class 'object'>, method_or_property=__call__```

It does not happen in the normal (not callable) protocol tests.

File: jsonargparse/_typehints.py (hunk 5 above)
Comment (MiguelMonteiro) [reply jsonargparse/_typehints.py:1110]: Shall I open an issue so we can keep track of this feature/bug?

Comment (mauvilsa) [reply jsonargparse/_typehints.py:1110]: Sure, open an issue.

File: jsonargparse/_typehints.py [hunk 6]
```diff
@@ -1087,15 +1087,32 @@ def adapt_typehints(
...
+    "__subclasshook__",
+}
+
+
 def implements_protocol(value, protocol) -> bool:
     from jsonargparse._parameter_resolvers import get_signature_parameters
     from jsonargparse._postponed_annotations import get_return_type
 
-    if not inspect.isclass(value):
+    if not inspect.isclass(value):  # Should we check if callables implement protocols?
```
mauvilsa, line 1110:
```suggestion
    if not inspect.isclass(value) or value is object:
```
Simple solution to avoid the problem with object.

jsonargparse_tests/test_subclasses.py
```diff
@@ -1500,6 +1500,66 @@ def test_parse_implements_protocol(parser):
...
+
+    def __call__(self, items: List[float]) -> List[float]:
+        return items
+
+
+# TODO: make functions that implement protocol be checked correctly
+def implements_callable_interface2(items: List[float]) -> List[float]:
+    return items
+
+
```
mauvilsa, line 1522:
```suggestion
```
Add this when it gets implemened.

jsonargparse_tests/test_subclasses.py
```diff
@@ -1500,6 +1500,66 @@ def test_parse_implements_protocol(parser):
...
+@pytest.mark.parametrize(
+    "expected, value",
+    [
+        (True, ImplementsCallableInterface1),
+        (False, ImplementsCallableInterface1(1)),
+        (
+            False,
+            implements_callable_interface2,
+        ),  # TODO: switch to True once functions that implement protocol are checked correctly
```
mauvilsa, line 1551:
```suggestion
```
Add this when it gets implemened.

mauvilsa (changes requested):
Looks good. But please remove the TODOs as commented.

MiguelMonteiro:
all should be good now

mauvilsa (approved):

PR #661, opened 2025-01-22
Experimental support for sub-classing ArgumentParser to customize add_argument

### LINKED ISSUE #301: Question: is it possible to make CLI replace parameter dashes with underscores?

### LINKED ISSUE #355: CLI: can bool parameters become `--flag` instead of `--flag true`?
## 🚀 Feature request

Currently, while using CLI(), a bool param must be specified on the command line as `--flag true` or `--flag yes`. Is it possible to set it up such that `--flag` without the `yes` or `true` will still pass `True` into the function?

### Motivation

I see this commonly tripping up members on my team. `--flag` is a common pattern, `store_true` in argparse, supported in python-fire (https://github.com/google/python-fire/blob/master/docs/guide.md#boolean-arguments)

### Pitch

`--flag` should work the same as `--flag yes` or `--flag true`

mauvilsa:
## What does this PR do?

Fixes #301 - change signature underscores to dashes
Fixes #355 - customize signature bool parameters

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,8 +18,8 @@ v4.37.0 (2025-01-??)
 Added
 ^^^^^
 - Experimental support for sub-classing ``ArgumentParser`` to customize
-  ``add_argument`` (`#???
-  <https://github.com/omni-us/jsonargparse/pull/???>`__).
+  ``add_argument`` (`#661
+  <https://github.com/omni-us/jsonargparse/pull/661>`__).
 
 
 v4.36.0 (2025-01-17)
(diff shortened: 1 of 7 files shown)


petergaultney:
i did some fairly rudimentary testing of my own and this seems to work. I appreciate you solving for the more general case that includes argument groups.

PR #665, opened 2025-01-30
feat: support for partialmethod

### LINKED ISSUE #664: Add support for partialmethods and partial types
## 🚀 Feature request

Add support for objects created via functools.partial and functools.partialmethod

### Motivation

Sometimes dynamically created classes and functions via partialmethod and partial are useful to prevent repeating code.

### Pitch

I was able to  implement this locally by changing the following:
Implement a function that checks if the method of class partial or partialmethod
```
from functools import partial, partialmethod
def is_partial_method(method):
    return isinstance(method, partialmethod) or isinstance(method, partial)
```
add another code path in `get_component_and_parent` in file `_parameter_resolver.py`:
```
def get_component_and_parent(
    function_or_class: Union[Callable, Type],
    method_or_property: Optional[Union[str, Callable]] = None,
):
    if is_subclass(function_or_class, ClassFromFunctionBase) and method_or_property in {None, "__init__"}:
        function_or_class = function_or_class.wrapped_function  # type: ignore[union-attr]
        if isinstance(function_or_class, MethodType):
            method_or_property = function_or_class.__name__
            function_or_class = function_or_class.__self__  # type: ignore[assignment]
        else:
            method_or_property = None
    elif inspect.isclass(get_generic_origin(function_or_class)) and method_or_property is None:
        method_or_property = "__init__"
    elif method_or_property and not isinstance(method_or_property, str):
        method_or_property = method_or_property.__name__
    parent = component = None
    if method_or_property:
        attr = inspect.getattr_static(get_generic_origin(function_or_class), method_or_property)
        if is_staticmethod(attr):
            component = getattr(function_or_class, method_or_property)
            return component, parent, method_or_property
        parent = function_or_class
        if has_dunder_new_method(function_or_class, method_or_property):
            component = getattr(function_or_class, "__new__")
        elif is_method(attr):
            component = attr
        elif is_partial_method(attr):
            component = getattr(function_or_class, method_or_property)
        elif is_property(attr):
            component = attr.fget
        elif isinstance(attr, classmethod):
            component = getattr(function_or_class, method_or_property)
        elif attr is not object.__init__:
            raise ValueError(
                f"Invalid or unsupported input: class={function_or_class}, method_or_property={method_or_property}"
            )
    else:
        if not callable(function_or_class):
            raise ValueError(f"Non-callable input: function={function_or_class}")
        component = function_or_class
    return component, parent, method_or_property

```

Just need to open and MR but thought I would create an issue first

MiguelMonteiro:
## What does this PR do?

Adds support for functools.partial and functools.partialmethod and closes https://github.com/omni-us/jsonargparse/issues/664
Still need to write tests and update documentation but would like wait for @mauvilsa to weigh in to see if the idea makes sense.

## Before submitting

- [ x] Did you read the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)?


mauvilsa (changes requested):
I can't think of a reason why not to support this. Please go ahead and write the unit tests.

MiguelMonteiro:
> I can't think of a reason why not to support this. Please go ahead and write the unit tests.

I was able to add support for `partialmethod` and added a test for it. For partial functions and classes it's a bit trickier since they are dynamic and it seems the parameter resolver is expecting static objects for the most part.

The options would be to add a subsection in `get_component_and_parent` that checks if the `function_or_class` is a partial and handle it separately? Or change the way the attr is obtained form statically `attr = inspect.getattr_static(get_generic_origin(function_or_class), method_or_property)` to dynamically. But I don't fully understand the downstream effects of these changes.

jsonargparse_tests/test_parameter_resolvers.py
```diff
@@ -899,6 +902,14 @@ def test_get_params_some_ignored():
         assert_params(get_params(func_given_kwargs), ["p", "p1"], help=False)
 
 
+# test partial method
```
mauvilsa, line 905: Why this comment? It is redundant, since it is the same as the function name.

```suggestion
```

CHANGELOG.rst
```diff
@@ -20,7 +20,8 @@ Added
 - Experimental support for sub-classing ``ArgumentParser`` to customize
   ``add_argument`` (`#661
   <https://github.com/omni-us/jsonargparse/pull/661>`__).
-
+- Support for partialmethods (`#664
+  <https://github.com/omni-us/jsonargparse/issues/664>`__).
```
mauvilsa, line 24: Please reference the pull request, not the issue.

```suggestion
- Support for partialmethods (`#665
  <https://github.com/omni-us/jsonargparse/pull/665>`__).
```

mauvilsa (changes requested):
> I was able to add support for partialmethod and added a test for it. For partial functions and classes it's a bit trickier since they are dynamic and it seems the parameter resolver is expecting static objects for the most part.

What was the reason you came to know that partials didn't work? I would guess it was because you needed it for some use case. If it was only for `partialmethod`, then what is implemented now already works and is a new feature. If you really want `partial` to also work, later I can take a look.

MiguelMonteiro:
> > I was able to add support for partialmethod and added a test for it. For partial functions and classes it's a bit trickier since they are dynamic and it seems the parameter resolver is expecting static objects for the most part.
> 
> What was the reason you came to know that partials didn't work? I would guess it was because you needed it for some use case. If it was only for `partialmethod`, then what is implemented now already works and is a new feature. If you really want `partial` to also work, later I can take a look.

For classes the main reason is that `partial` returns an instance of `<class 'functools.partial'>` which is not a subclass of the original class. This breaks type hints and detecting subclasses as far as I could tell. Additionally, for both Callables and classes, partial returns a descriptor, hiding the wrapped object under `__get__`,. This requires changes to be made in the parameter resolution at least, but I can't tell if there are other places which would need to be changed to handle descriptors.

For reference, my use case was to create partial classes. I got around the limitations by using `partialmethod` and explicitly creating a new type (instead of the original type wrapped in partial):
```
cls = type(
        name,
        (base_class,),
        {
            "__init__": partialmethod(base_class.__init__, **keywords),
            "__doc__": base_class.__doc__,
        },
    )
   ```

File: jsonargparse_tests/test_parameter_resolvers.py (hunk 1 above)
Comment (MiguelMonteiro) [reply jsonargparse_tests/test_parameter_resolvers.py]: sure, removed the comment

File: jsonargparse_tests/test_parameter_resolvers.py [hunk 3]
```diff
@@ -899,6 +902,13 @@ def test_get_params_some_ignored():
         assert_params(get_params(func_given_kwargs), ["p", "p1"], help=False)
 
 
+def test_partialmethod():
+    ClassA.partial_method_a = partialmethod(ClassA.method_a, pma1=1, pma2=0.5)
+    assert_params(get_params(ClassA, "partial_method_a"), ["pma1", "pma2", "kma1"])
```
mauvilsa, line 907: Looking again at this test now I think there are things missing. The partial doesn't change the list of parameter names. But it does change the defaults. Are the defaults correctly inferred by a parser? If not, then it doesn't really work as expected. So this should be tested.

And another thing that could be considered missing is that a method could be set as default to a callable. For this `--print_config` should serialize this default into an import path string. And this import path when used in a config, should be correctly deserialize into the partial method object.

mauvilsa:
For partial functions and classes being dynamic, the import path would also be a problem. If a partial function is saved at the top level of a module in a variable, from the partial function object alone can it be determined what its import path should be?

mauvilsa:
I updated the branch because the checks were not working correctly.

mauvilsa (approved):

Issue #685, opened 2025-02-26
"Providing both type and action not allowed" check too strict


PabloLION:
## 🐛 Bug report

1. When we use both argument `action=None` and `type=None` in `add_argument()`, raises "Providing both type and action not allowed". See the repro steps below.

2. (BTW) Documentation is not clear about the differences between `jsonargparse` and `argparse`.

   - In the docs, the [`ActionsContainer.add_argument()`](https://jsonargparse.readthedocs.io/en/v4.37.0/index.html#jsonargparse.ActionsContainer.add_argument) method of `ArgumentParser` instance behaves slightly different than the `argparse.ArgumentParser.add_argument()` method. However, it's not pointed out in the docs. The `jsonargparse` docs say:

     > Adds an argument to the parser or argument group.
     > All the arguments from argparse.ArgumentParser.add_argument are supported. Additionally it accepts:
     > Parameters: enable_path ...

   - When we use it, with `jsonargparse`, the `add_argument()` method raises "Providing both type and action not allowed". This is not the case in `argparse` with Python (3.13.1). I understand this is a design choice, but it is not documented in the API reference.

### To reproduce

1. Using `jsonargparse.ArgumentParser`, with jsonargparse v4.37.0.

    ```python
    import jsonargparse

    parser = jsonargparse.ArgumentParser(description="CLI with config integration.")

    # action with type=None is ok.
    parser.add_argument(
        "--bool",
        type=None,
        action=jsonargparse.ActionYesNo,
        default=False,
        help="example bool",
    )

    # Cannot add args: ValueError: Providing both type and action not allowed.
    parser.add_argument(
        "--string",
        type=str,
        action=None,
        default="default",
        help="example string",
    )
    ```

2. run the code above, it will show the error:

    ```shell
    ValueError: Providing both type and action not allowed
    ```

### Expected behavior

Allow user to pass in `action=None` as if it was not used.
My current work around is ugly:

```python
if action is None:
    parser.add_argument(
        arg_name,
        type=arg_type,
        default=arg_default,
        help=help_text,
    )
elif arg_type is None:
    parser.add_argument(
        arg_name,
        action=action,
        default=arg_default,
        help=help_text,
    )
else:
    raise ValueError(
        f"Cannot have both action and arg_type. For {arg_name=}, {action=}, {arg_type=}"
    )
```

### Additional context

[...]

PR #687, opened 2025-03-03
Fix/type action restriction

PabloLION:
## What does this PR do?

Fix https://github.com/omni-us/jsonargparse/issues/685
Now in `add_argument`, passing `action=None` behaves like call without argument `action`.

```py3
parser.add_argument(
    "--string",
    type=str,
    action=None,
    default="default",
    help="example string",
)
```

## Before submitting

### DIFF
--- a/jsonargparse/_core.py
+++ b/jsonargparse/_core.py
@@ -122,7 +122,7 @@ class ActionsContainer(SignatureArguments, argparse._ActionsContainer):
             enable_path: Whether to try parsing path/subconfig when argument is a complex type.
         """
         parser = self.parser if hasattr(self, "parser") else self
-        if "action" in kwargs:
+        if "action" in kwargs and kwargs["action"] is not None:
             if ActionParser._is_valid_action_parser(parser, kwargs["action"]):
                 return ActionParser._move_parser_actions(parser, args, kwargs)
             ActionConfigFile._ensure_single_config_argument(self, kwargs["action"])
--- a/jsonargparse/_core.py
+++ b/jsonargparse/_core.py
@@ -122,7 +122,7 @@ class ActionsContainer(SignatureArguments, argparse._ActionsContainer):
             enable_path: Whether to try parsing path/subconfig when argument is a complex type.
         """
         parser = self.parser if hasattr(self, "parser") else self
-        if "action" in kwargs and kwargs["action"] is not None:
+        if kwargs.get("action") is not None:
             if ActionParser._is_valid_action_parser(parser, kwargs["action"]):
                 return ActionParser._move_parser_actions(parser, args, kwargs)
             ActionConfigFile._ensure_single_config_argument(self, kwargs["action"])
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -280,7 +280,7 @@ class ActionTypeHint(Action):
 
     @staticmethod
     def prepare_add_argument(args, kwargs, enable_path, container, logger, sub_add_kwargs=None):
-        if "action" in kwargs and kwargs["action"] is not None:
+        if kwargs.get("action") is not None:
             raise ValueError("Providing both type and action not allowed.")
         typehint = kwargs.pop("type")
         if args[0].startswith("--") and ActionTypeHint.supports_append(typehint):
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -280,7 +280,7 @@ class ActionTypeHint(Action):
 
     @staticmethod
     def prepare_add_argument(args, kwargs, enable_path, container, logger, sub_add_kwargs=None):
-        if "action" in kwargs:
+        if "action" in kwargs and kwargs["action"] is not None:
             raise ValueError("Providing both type and action not allowed.")
         typehint = kwargs.pop("type")
         if args[0].startswith("--") and ActionTypeHint.supports_append(typehint):
--- a/jsonargparse_tests/test_typehints.py
+++ b/jsonargparse_tests/test_typehints.py
@@ -68,6 +68,20 @@ def test_add_argument_failure_given_type_and_action(parser):
     ctx.match("Providing both type and action not allowed")
 
 
+def test_add_argument_given_type_and_null_action(parser):
+    parser.add_argument("--op1", type=Optional[bool], action=None)
+    assert parser.get_defaults().op1 is None
+
+
+@parser_modes
+def test_add_argument_str_edge_cases_type_and_null_action(parser):
+    parser.add_argument("--val", type=str, action=None)
+    assert parser.parse_args(["--val=e123"]).val == "e123"
+    assert parser.parse_args(["--val=123e"]).val == "123e"
+    val = "1" * 5000
+    assert parser.parse_args([f"--val={val}"]).val == val
+
+
 @pytest.mark.parametrize("typehint", [Namespace, Optional[Namespace], Union[int, Namespace], List[Namespace]])
 def test_namespace_unsupported_as_type(parser, typehint):
     with pytest.raises(ValueError, match="Namespace .* not supported as a type"):
--- a/jsonargparse_tests/test_typehints.py
+++ b/jsonargparse_tests/test_typehints.py
@@ -73,15 +73,6 @@ def test_add_argument_given_type_and_null_action(parser):
     assert parser.get_defaults().op1 is None
 
 
-@parser_modes
-def test_add_argument_str_edge_cases_type_and_null_action(parser):
-    parser.add_argument("--val", type=str, action=None)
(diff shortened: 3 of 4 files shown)


PabloLION:
## Additional info
When creating remote branch, the git hook fired and I got this

```sh
mypy................................................................................Failed
- hook id: mypy
- duration: 0.98s
- exit code: 1

jsonargparse/_stubs_resolver.py:51: error: No overload variant of "Assign" matches argument types "list[Name | Attribute | Subscript]", "expr | None", "list[Never]", "int", "int"  [call-overload]
jsonargparse/_stubs_resolver.py:51: note: Possible overload variants:
jsonargparse/_stubs_resolver.py:51: note:     def Assign(self, targets: list[expr], value: expr, type_comment: str | None = ..., *, lineno: int = ..., col_offset: int = ..., end_lineno: int | None = ..., end_col_offset: int | None = ...) -> Assign
jsonargparse/_stubs_resolver.py:51: note:     def Assign(self, targets: list[expr] = ..., *, value: expr, type_comment: str | None = ..., lineno: int = ..., col_offset: int = ..., end_lineno: int | None = ..., end_col_offset: int | None = ...) -> Assign
jsonargparse/_stubs_resolver.py:53: error: Unused "type: ignore" comment  [unused-ignore]
jsonargparse/_stubs_resolver.py:54: error: Unused "type: ignore" comment  [unused-ignore]
Found 3 errors in 1 file (checked 3 source files)
```

Since I didn't edit these files, I used `--no-verify`.

Also, I didn't have mypy/ruff installed (which made. Maybe we can consider adding them to the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)? I saw them in `pyproject.toml` but I could format my code :-( without them.

PabloLION:
not sure where to put the test:

- test_typehints (I added here: the edited file is _typehints.py, and the feature is also about a valid `type` with a `None` action)
- test_actions (the feature is about `action`)
- test_core (the feature is about `add_argument`)
- test_formatters (do we need to test how help changes? like adding "test_help_type_and_no_action")

PabloLION:
Also I find another problem: we pass in `choices=None` it works differently here
https://github.com/omni-us/jsonargparse/blob/c3d3df7bee6b606616ee8f3ac41d09a05a2ac90d/jsonargparse/_core.py#L142

Should I create another PR on this?

jsonargparse/_core.py
```diff
@@ -122,7 +122,7 @@ def add_argument(self, *args, enable_path: bool = False, **kwargs):
             enable_path: Whether to try parsing path/subconfig when argument is a complex type.
         """
         parser = self.parser if hasattr(self, "parser") else self
-        if "action" in kwargs:
+        if "action" in kwargs and kwargs["action"] is not None:
```
mauvilsa, line 125:
```suggestion
        if kwargs.get("action") is not None:
```

jsonargparse/_typehints.py
```diff
@@ -280,7 +280,7 @@ def normalize_default(self, default):
 
     @staticmethod
     def prepare_add_argument(args, kwargs, enable_path, container, logger, sub_add_kwargs=None):
-        if "action" in kwargs:
+        if "action" in kwargs and kwargs["action"] is not None:
```
mauvilsa, line 283:
```suggestion
        if kwargs.get("action") is not None:
```

jsonargparse_tests/test_typehints.py
```diff
@@ -68,6 +68,20 @@ def test_add_argument_failure_given_type_and_action(parser):
...
+
+@parser_modes
+def test_add_argument_str_edge_cases_type_and_null_action(parser):
+    parser.add_argument("--val", type=str, action=None)
+    assert parser.parse_args(["--val=e123"]).val == "e123"
+    assert parser.parse_args(["--val=123e"]).val == "123e"
+    val = "1" * 5000
+    assert parser.parse_args([f"--val={val}"]).val == val
+
+
```
mauvilsa, line 84:
```suggestion
```

Why this test? Isn't `test_add_argument_given_type_and_null_action` enough to cover the case of action being `None`?

mauvilsa (changes requested):
Issue #685 also mentioned the case `type=None` so I was expecting a test for this.

mauvilsa:
> When creating remote branch, the git hook fired and I got this
> 
> ```shell
> mypy................................................................................Failed
> ```

I see. I think I need to specify in the pre-commit config which python version to run mypy with. No problem that you skipped it.

> Also, I didn't have mypy/ruff installed (which made. Maybe we can consider adding them to the [contributing guideline](https://github.com/omni-us/jsonargparse/blob/main/CONTRIBUTING.rst)? I saw them in `pyproject.toml` but I could format my code :-( without them.

mypy/ruff don't need to be installed. It is `pre-commit` the one that installs them in a separate environment. If you want to only run `ruff` do
```shell
pre-commit run -a ruff
```
to only run `mypy`, do
```shell
pre-commit run -a --hook-stage pre-push mypy
```

jsonargparse_tests/test_typehints.py
PabloLION: Yes, it is enough. I wanted to make sure weird things won't happen but this is not the correct way. It makes a lot of sense to not add it.

PabloLION: With `type=None` and `action=None`, there's no problem with the old code. Should we still add the test?

PabloLION:
> > Also I find another problem: we pass in `choices=None` it works differently here
> > https://github.com/omni-us/jsonargparse/blob/c3d3df7bee6b606616ee8f3ac41d09a05a2ac90d/jsonargparse/_core.py#L142
> > 
> > Should I create another PR on this?
> 
> How does it work differently? I think it could be part of this pull request.

When `action is not None`, passing `choices=None`, there will be similar problem with `type=None`.
But I have more doubts about choices: I'm creating a `parser` with `dataclass` where some fields are Enum type, and only a part of the Enum should be used. (Enum CLI args are matched by the name instead of value.)
In this case, jsonargparse will check if the parsed value is in `choices`, not the user input. However, the help test is only showing the parsed option, not what user should input.

I meant to create a new issue but I have not read the document nor have I tested it yet. Do you think this worths a shot?

mauvilsa (changes requested):
Please add an entry to the changelog. After that, this would be good to merge.

> With type=None and action=None, there's no problem with the old code. Should we still add the test?

Okay, fine. Then lets leave it like that.

> When action is not None, passing choices=None, there will be similar problem with type=None.
> But I have more doubts about choices: I'm creating a parser with dataclass where some fields are Enum type, and only a part of the Enum should be used. (Enum CLI args are matched by the name instead of value.)
> In this case, jsonargparse will check if the parsed value is in choices, not the user input. However, the help test is only showing the parsed option, not what user should input.
>
> I meant to create a new issue but I have not read the document nor have I tested it yet. Do you think this worths a shot?

Okay, it does sound like this needs a specific issue to address.

PabloLION:
I added the change log and I think it's ready to merge.

> Can you explain what you mean by "I cannot format the code effectively"? The code will be automatically formatted when pre-commit runs. And even if pre-commit is skipped, a github action will format it.

I wanted to say that I avoided formatting the code intentionally because my line length was set to 80 instead of 120 on my local environment because I didn't know about the GitHub action. And Pylance was reporting this everywhere

```text
"_ActionPrintConfig" is private and used outside of the module in which it is declared (Pylance) 
[reportPrivateUsage](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportPrivateUsage)
```

And about the Union type deprecation "This type is deprecated as of Python 3.10" etc. I guess our env was very different and I didn't change my python interpreter version (I'm exactly suffering for "mac313").

Anyway, thanks for your patience and guidance in this PR. I'll create another issue on the "choices" args.

mauvilsa (approved):
Thank you for contributing!

mauvilsa:
> > Can you explain what you mean by "I cannot format the code effectively"? The code will be automatically formatted when pre-commit runs. And even if pre-commit is skipped, a github action will format it.
> 
> I wanted to say that I avoided formatting the code intentionally because my line length was set to 80 instead of 120 on my local environment because I didn't know about the GitHub action. And Pylance was reporting this everywhere
> 
> ```
> "_ActionPrintConfig" is private and used outside of the module in which it is declared (Pylance) 
> [reportPrivateUsage](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportPrivateUsage)
> ```
> 
> And about the Union type deprecation "This type is deprecated as of Python 3.10" etc. I guess our env was very different and I didn't change my python interpreter version (I'm exactly suffering for "mac313").

Thanks for the info! I will do some changes and probably ping you for review.

PR #692, opened 2025-03-11
Support parsing optionals as positionals

### LINKED ISSUE #690: enable all arguments as positional
## 🚀 Feature request

Following the very first code sample in docs and wondering if it is possible to relax the argument parsing and enable to pass all arguments as positional

### Motivation

Limit some unexpected crashes when passing all arguments about some without its name

```py
from jsonargparse import CLI


def command(name: str, prize: int = 100):
    """Prints the prize won by a person.

    Args:
        name: Name of winner.
        prize: Amount won.
    """
    print(f"{name} won {prize}€!")


if __name__ == "__main__":
    CLI(command)
```

and call it as
```bash
python scripts/trying.py any 200
```

gives following:
```
usage: trying.py [-h] [--config CONFIG] [--print_config[=flags]] [--prize PRIZE] name
error: Unrecognized arguments: 200
```

### Pitch

This prevents us from switching from Fire, and this would be breaking many of our workflows
ref: https://github.com/Lightning-AI/utilities/pull/371

mauvilsa:
## What does this PR do?

Fixes #690

## Before submitting

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,8 @@ Added
 ^^^^^
 - Support ``shtab`` completion of ``Literal`` types (`#693
   <https://github.com/omni-us/jsonargparse/pull/693>`__).
+- Support for parsing optionals as positionals (`#692
+  <https://github.com/omni-us/jsonargparse/pull/692>`__).
 
 Changed
 ^^^^^^^
--- a/DOCUMENTATION.rst
+++ b/DOCUMENTATION.rst
@@ -332,6 +332,51 @@ This can be easily implemented with :func:`.capture_parser` as follows:
     :func:`.auto_cli` is by using :func:`.capture_parser`.
 
 
+Optionals as positionals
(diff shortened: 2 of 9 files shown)


Borda:
could we bump the version to an RC or something so I can test it with our CI - https://github.com/Lightning-AI/utilities/pull/371

mauvilsa:
> could we bump the version to an RC or something so I can test it with our CI - [Lightning-AI/utilities#371](https://github.com/Lightning-AI/utilities/pull/371)

This looks good https://github.com/Lightning-AI/utilities/blob/9610d5bfa0a639d0820ef525e5cec06b0db71e11/requirements/cli.txt#L1, I think there is no need for an RC.

mauvilsa (commented):
I have added more tests and documentation for the feature. But it is still pending to not support this for subclass and dataclass types. What to do regarding shtab can be a future improvement.

PR #758, opened 2025-08-15
Add support callable protocols for instance factory dependency injection

### LINKED ISSUE #755: Allow partial application with keyword arguments
## 🚀 Feature request
The current implementation of partial application for callables as described [here](https://jsonargparse.readthedocs.io/en/v4.40.2/#callable-type) only allows for positional arguments.
 Would it be possible to extend the partial application support to allow for keyword arguments?
 
### Motivation
After our discussion in https://github.com/omni-us/jsonargparse/issues/745 I started using partial application to instantiate objects after dependency injection, but I started noticing a few problems.

Since the arguments of callable types in code are unnamed, they don't document what each argument means semantically: 
In `dataloader: Callable[[Dataset, int], DataLoader]` the `int` is the `batch_size`, but thats not visible in the code.

For constructors / functions of library code, I regularly had to write wrappers to reorder arguments, so they begin with the arguments that aren't available at injection time.
Let's say, I want to provide the `dataset` and `shuffle` arguments of `DataLoader` after injection, I have to create something like this:
```py
class DataModule(LightningDataModule):
    def __init__(self, dataloader_factory: Callable[[Dataset, bool], torch.utils.data.DataLoader]):
        pass

class DataloaderWrapper(torch.utils.data.DataLoader):
    def __init__(self, dataset: Dataset, shuffle: bool, **kwargs):
        super().__init__(
            dataset=dataset,
            shuffle=shuffle,
            **kwargs
        )

# or
def dataloader_factory(dataset: Dataset, shuffle: bool, **kwargs):
    return torch.utils.data.DataLoader(
        dataset=dataset,
        shuffle=shuffle,
        **kwargs
    )
```

### Pitch
Would it be possible to annotate the callable with `typing.Protocol` as described in the [Python documentation](https://docs.python.org/3.13/library/typing.html#annotating-callable-objects) and have jsonargparse translate this to a partial application with keyword arguments?
I understand that `Protocol` cannot be used directly here since it is supported by jsonargparse to constrain types, but a wrapper should be possible, right?

```py
class DataloaderFactory(jsonargparse.CallableProtocol):
    def __call__(self, dataset: Dataset, shuffle: bool) -> DataLoader: ...

class DataModule(LightningDataModule):
    def __init__(self, dataloader_factory: DataloaderFactory):
        pass
```

This way, DataloaderFactory can be defined right next to the DataModule and the signature including argument names is easily visible as well as checkable by jsonargparse and any static type checker.

mauvilsa:
## What does this PR do?

Fixes #755

## Before submitting

- [x] Did you update **the CHANGELOG** with the pull request link? (not for typos, docs, test updates, or minor internal changes/refactors)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,8 @@ Added
 ^^^^^
 - Support for Python 3.14 (`#753
   <https://github.com/omni-us/jsonargparse/pull/753>`__).
+- Support callable protocols for instance factory dependency injection (`#758
+  <https://github.com/omni-us/jsonargparse/pull/758>`__).
 
 Fixed
 ^^^^^
--- a/DOCUMENTATION.rst
+++ b/DOCUMENTATION.rst
@@ -1035,96 +1035,9 @@ A second option is a class that once instantiated becomes callable:
     >>> init.callable(5)
     8
 
-The third option is only applicable when the type is a callable that has a class
-as return type or a ``Union`` including a class. This is useful to support
-dependency injection for classes that require a parameter that is only available
-after injection. The parser supports this automatically by providing a function
-that receives this parameter and returns the instance of the class. Take for
-example the classes:
-
-.. testcode:: callable
-
-    class Optimizer:
-        def __init__(self, params: Iterable):
-            self.params = params
-
-
-    class SGD(Optimizer):
-        def __init__(self, params: Iterable, lr: float):
-            super().__init__(params)
-            self.lr = lr
-
-.. testcode:: callable
-    :hide:
-
-    doctest_mock_class_in_main(SGD)
-
-A possible parser and callable behavior would be:
-
-.. doctest:: callable
-
-    >>> value = {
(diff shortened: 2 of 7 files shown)


mauvilsa:
@E1k3 please try this out to help identify issues and possible edge cases. Later I will probably add more tests and will update the documentation.

mauvilsa:
@E1k3 I have now added documentation changes and one more test.

mauvilsa:
Merging this. Anyway people can comment here, and if there is a need to fix, then it can be done in a separate pull request.

E1k3:
> @E1k3 please try this out to help identify issues and possible edge cases. Later I will probably add more tests and will update the documentation.

Wow that was fast. I was on vacation until today, but I will use this feature in my code and also maybe try some more obscure use cases and let you know if I run into any problems.

Thank you for your work! Jsonargparse might be the best part about working with Lightning for me and will probably also replace standard argparse in most of my personal projects in the future.

PR #780, opened 2025-10-04
Minor patch to support pydantic basemodel better

findmyway:
This is to make https://github.com/omni-us/jsonargparse/issues/778#issuecomment-3367797529 work.


jsonargparse/_core.py
```diff
@@ -1155,7 +1155,7 @@ def check_values(cfg):
                     if group_key:
                         subkey = key[len(group_key) + 1 :]
                         raise NSKeyError(f"Group '{group_key}' does not accept nested key '{subkey}'")
-                    raise NSKeyError(f"Key '{key}' is not expected")
+                    # raise NSKeyError(f"Key '{key}' is not expected")
```
findmyway, line 1158: Still not very sure how to disable it.

Issue #904, opened 2026-04-24
_typehints.py: 5 unhandled-edge-case crashes (forward refs in __main__, bare tuple/set, Dict[int, str] with bad keys)


pramodavansaber:
Hi! Found five edge-case crashes in `jsonargparse/_typehints.py` while running adversarial test generation against the repo via [tailtest](https://tailtest.com). All five raise unhandled exceptions on legitimate inputs that should produce a controlled error or a typed result.

### 1. `resolve_forward_ref` crashes when `__builtins__` is a module (line 777)

```python
__builtins__.copy()  # AttributeError when __builtins__ is the module, not the dict
```

`__builtins__` is the `builtins` module in `__main__` and many embedding contexts; in regular modules it's a dict. Both should work.

**Fix:** `import builtins; vars(builtins).copy()`

### 2 & 3. `is_ellipsis_tuple()` AttributeError on bare `tuple` / `set` types (line 1633)

Bare `tuple` and `set` (without subscripting) have no `__origin__`. The function accesses it unconditionally → `AttributeError`.

**Fix:** `getattr(typehint, "__origin__", None)` or use the existing `get_typehint_origin()` helper.

### 4 & 5. `Dict[int, str]` with non-numeric / empty-string keys leaks raw `ValueError` (line 972)

Casting user-supplied JSON keys to int via `int(k)` raises bare `ValueError("invalid literal for int() with base 10: '...'")`. Users see the raw Python error instead of a typed message tied to the offending field.

**Fix:** wrap in try/except and raise a `TypeError`/`ValueError` with the schema context: which key, which type was expected.

### Reproduction

```python
import pytest
from typing import Dict


# Bug 1
def test_resolve_forward_ref_when_builtins_is_module():
    """resolve_forward_ref crashes if __builtins__ is the module (e.g. in __main__)."""
    import builtins, jsonargparse._typehints as th
    # Simulate the __main__ context where __builtins__ is the module
    fake_globals = {"__builtins__": builtins}
    try:
        th.resolve_forward_ref("int", aliases=fake_globals)
    except AttributeError as e:
        if "module 'builtins' has no attribute 'copy'" in str(e):
            pytest.fail(f"BUG: __builtins__ as module crashes: {e}")


# Bugs 2 & 3
def test_is_ellipsis_tuple_bare_tuple():
    from jsonargparse._typehints import is_ellipsis_tuple
    try:
        is_ellipsis_tuple(tuple)  # bare tuple type
[...]

PR #911, opened 2026-05-11
fix(typehints): handle bare tuple/set and module-form __builtins__

SAY-5:
## What does this PR do?

Fixes two of the crashes reported in #904.

`is_ellipsis_tuple` raised `AttributeError` when called on bare tuple/set/list types (no subscript) because it accessed `__origin__` unconditionally. Switched to `getattr` with a default so bare types return False.

`resolve_forward_ref` called `__builtins__.copy()`, which fails in `__main__` and many embedding contexts where `__builtins__` is the builtins module rather than its dict. Added an isinstance check to handle both forms.

Added parametrized tests covering bare types and subscripted tuple variants in test_typehints.py.

## Before submitting

- [x] Did you read the contributing guideline?
- [x] Did you write unit tests?
- [x] Did you verify that new and existing tests pass locally?
- [x] Did you make sure that all changes preserve backward compatibility?

### DIFF
--- a/jsonargparse_tests/test_typehints.py
+++ b/jsonargparse_tests/test_typehints.py
@@ -1657,6 +1657,18 @@ def test_is_ellipsis_tuple_with_subscripted_tuple():
     assert is_ellipsis_tuple(tuple[int, ...]) is True
 
 
+def test_argument_parser_accepts_bare_tuple_typehint(parser):
+    parser.add_argument("--data", type=tuple)
+    cfg = parser.parse_args(["--data=[1, 2]"])
+    assert tuple(cfg.data) == (1, 2)
+
+
+def test_argument_parser_accepts_bare_set_typehint(parser):
+    parser.add_argument("--data", type=set)
+    cfg = parser.parse_args(["--data=[1, 2, 1]"])
+    assert set(cfg.data) == {1, 2}
+
+
 class SkipDefault(Calendar):
     def __init__(self, *args, param: str = "0", **kwargs):
         super().__init__(*args, **kwargs)
--- a/jsonargparse_tests/test_typehints.py
+++ b/jsonargparse_tests/test_typehints.py
@@ -1642,6 +1642,21 @@ def test_is_optional(typehint, ref_type, expected):
     assert expected == is_optional(typehint, ref_type)
 
 
+@pytest.mark.parametrize("bare_type", [tuple, set, list, frozenset])
+def test_is_ellipsis_tuple_accepts_bare_types(bare_type):
+    from jsonargparse._typehints import is_ellipsis_tuple
+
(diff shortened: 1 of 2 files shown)


jsonargparse_tests/test_typehints.py
```diff
@@ -1642,6 +1642,33 @@ def test_is_optional(typehint, ref_type, expected):
...
+
+    assert is_ellipsis_tuple(bare_type) is False
+
+
+def test_is_ellipsis_tuple_with_subscripted_tuple():
+    from jsonargparse._typehints import is_ellipsis_tuple
+
+    assert is_ellipsis_tuple(Tuple[int, ...]) is True
+    assert is_ellipsis_tuple(Tuple[int, str]) is False
+    assert is_ellipsis_tuple(tuple[int, ...]) is True
```
mauvilsa, line 1657: `is_ellipsis_tuple` is not public API. There is no need to change `is_ellipsis_tuple` if failures never happen when using the public API. I would need to see a test that uses the public API, and fails without the code fix.

jsonargparse_tests/test_typehints.py
```diff
@@ -1642,6 +1642,33 @@ def test_is_optional(typehint, ref_type, expected):
...
+
+
+def test_argument_parser_accepts_bare_tuple_typehint(parser):
+    parser.add_argument("--data", type=tuple)
+    cfg = parser.parse_args(["--data=[1, 2]"])
+    assert tuple(cfg.data) == (1, 2)
```
mauvilsa, line 1663: There is already a test for a bare tuple: `test_tuple_without_arg`. Also, if we undo the changes in `_typehint.py` this test passes. Which means, the changes in the code are not fixing anything.

jsonargparse_tests/test_typehints.py
```diff
@@ -1642,6 +1642,33 @@ def test_is_optional(typehint, ref_type, expected):
...
+
+
+def test_argument_parser_accepts_bare_set_typehint(parser):
+    parser.add_argument("--data", type=set)
+    cfg = parser.parse_args(["--data=[1, 2, 1]"])
+    assert set(cfg.data) == {1, 2}
```
mauvilsa, line 1669: If we undo the changes in `_typehint.py` this test passes. Which means, the changes in the code are not fixing anything. Though, there is no test for bare set, so this test could be kept.

mauvilsa (commented):
@SAY-5 thank you very much for your effort to contribute! However, looking at the changes I believe nothing is being fixed. As mentioned in #904 the findings are LLM generated, and they don't consider what is possible by using the public API. The added tests are either not for the public API, or don't fail without the code changes, so no fixes needed.

PR #942, opened 2026-07-31
fix: discard previous class defaults on subclass class_path change

### LINKED ISSUE #460: Instantiation of custom class leads to wrong defaults
## 🐛 Bug report

When instantiating objects from a custom class where a default is specified in the signature, jsonargparse infers the defaults from the the wrong class:

### To reproduce

```python
from jsonargparse import CLI


class Super:
    pass


class Default(Super):
    def __init__(self, foo: str = "foo", bar: str = "bar"):
        self.foo = foo
        self.bar = bar


class Other(Super):
    def __init__(self, foo: str = "foo", bar: str = "baaaaar"):
        self.foo = foo
        self.bar = bar


def main(data: Super = Default()):
    print(data.foo)
    print(data.bar)


if __name__ == "__main__":
    CLI(main)
```

```
python repro.py --data Other --data.foo test
```

Output:

```
test
bar
```

### Expected behavior

I expect the output to be

```
test
baaaaar
```

because I selected `--data Other`, so it should use the defaults from the `Other` class.

### Environment

- jsonargparse version (e.g., 4.8.0): 4.27.5
- Python version (e.g., 3.9): 3.10
- How jsonargparse was installed (e.g. `pip install jsonargparse[all]`): `pip install -U jsonargparse[signatures]
`
- OS (e.g., Linux): MacOS and Linux

Sanjays2402:
## What does this PR do?

Fixes #460. When a subclass argument has a default instance, its init_args were carried over to a newly selected class as long as they were valid for it. For init args that are merely the previous class's defaults, that shadowed the new class's own defaults, so `--data Other --data.foo test` still printed `Default`'s `bar` value. Init args whose value equals the previous class default and whose new class default differs are now discarded; explicitly given values are unaffected since they no longer equal the previous default.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

## Before submitting

- [x] Did you read the [contributing guideline](https://github.com/mauvilsa/jsonargparse/blob/main/CONTRIBUTING.rst)?
- [x] If you used a **coding agent**, did you fully understand and validate all generated code and ensure it follows the contributing guidelines?
- [ ] Did you update **the documentation**? (readme and public docstrings) — n/a, no public API change
- [x] If this is a **bug fix**, did you verify that the **tests fail without the code fix**?
- [x] Did you update **the CHANGELOG** including a pull request link? (not for typos, docs, test updates, or minor internal changes/refactors)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,10 @@ Fixed
 - ``TypedDict`` keys inherited from a base defined in a different module not
   resolving the types only imported in the ``TYPE_CHECKING`` block of that
   module (`#936 <https://github.com/mauvilsa/jsonargparse/pull/936>`__).
+- Init args of a default subclass instance not being discarded on class path
+  change when they are only the previous class defaults, thereby shadowing the
+  defaults of the newly selected class (`#460
+  <https://github.com/mauvilsa/jsonargparse/issues/460>`__).
 
 Changed
 ^^^^^^^
--- a/jsonargparse/_typehints.py
+++ b/jsonargparse/_typehints.py
@@ -1765,6 +1765,16 @@ def dump_kwargs_context(kwargs):
     yield
 
 
+def get_class_defaults(class_path, parser_or_action) -> Namespace:
+    if not class_path:
+        return Namespace()
+    sub_add_kwargs = getattr(parser_or_action, "sub_add_kwargs", {})
+    try:
+        return ActionTypeHint.get_class_parser(class_path, sub_add_kwargs).get_defaults()
+    except Exception:
+        return Namespace()
+
+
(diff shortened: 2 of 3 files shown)


mauvilsa (commented):
Thank you for your desire to contribute! Though, this can't be merged. As mentioned in https://github.com/mauvilsa/jsonargparse/issues/460#issuecomment-1970000917, this is the intended behavior. This pull request would just make the behavior inconsistent. The reason why #460 is open is because there could be a new feature to change this behavior for the parsers throughout. My last comments were me clarifying ideas to see if the overall behavior would make sense, which it didn't yet. Still, this would be a complex change not appropriate for a first time contributor.

